PR #47, opened 2019-04-26
fix(AEROGEAR-9093) - Error: configmaps mss-config not found and Probes

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9093

## What
* Fix the bug which was faced after a while because the DB was not founding the configMap with the values for env variables. 
* Improve and fix the probes for the DB and Service
 
## Why
Fix issues and probes to work with centos/redhat db image. 

## How
Ensure that all places will be looking for the correct name.

## Verification Steps
1. Clone the project, set up the Minishift, set up the mobile-security-service CR with the master VIP.  
2. Use the dev image `cmacedo/mobile-security-service-operator:AEROGEAR-9093-dev`. For it updates the `operator.yaml` file with this value
3. Run `make create-all` and check if all will be started with success. Also, check if the DB env variables are getting the values from the config map created to share it `mss-config`
4. Leave the instance running for a while, like 30 minutes,  then check if all still ok and if the env variables still using the ConfigMap values and no error as follows will be faced.

Screenshot 2019-04-26 at 10 24 20

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

Screenshot 2019-04-26 at 16 20 37

### DIFF
--- a/pkg/controller/mobilesecurityservicedb/helpers.go
+++ b/pkg/controller/mobilesecurityservicedb/helpers.go
@@ -18,7 +18,7 @@ func (r *ReconcileMobileSecurityServiceDB) getDatabaseNameEnvVar(m *mobilesecuri
 			ValueFrom: &corev1.EnvVarSource{
 				ConfigMapKeyRef: &corev1.ConfigMapKeySelector{
 					LocalObjectReference: corev1.LocalObjectReference{
-						Name: m.Spec.ConfigMapName,
+						Name: getConfigMapName(m),
 					},
 					Key: "PGDATABASE",
 				},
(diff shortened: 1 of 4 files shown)


pkg/controller/mobilesecurityservicedb/helpers.go
```diff
@@ -82,4 +82,12 @@ func (r *ReconcileMobileSecurityServiceDB) getDatabasePasswordEnvVar(m *mobilese
...
 		Value: m.Spec.DatabasePassword,
 	}
+}
+
+//getConfigMapName returns an string name with the name of the configMap
+func getConfigMapName(m *mobilesecurityservicev1alpha1.MobileSecurityServiceDB) string{
+	if len(m.Spec.ConfigMapName) > 0 {
+		return m.Spec.ConfigMapName
+	}
+	return m.Name
```
davidffrench: Does the name default to `mss-config`?

camilamacedo86: Hi @davidffrench, 

The name can be specified in the CR Spec as described in the README and on it.  So, if the value is specified then it will be the value used (m.Spec.ConfigMapName). 

We added a name/spec by default but it could be changed in the future. If no name is specified then it will use the instance name.

camilamacedo86:
Really tks for your help and support here @davidffrench.

PR #73, opened 2019-05-10
feat(finalizer) : Following the recommendations over how to work with finalizers

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9161
## What
* Remove the finalizer metadata from the YAML file and add it in the reconcile process

## Why
* Following the operator-framework recommendations

## Verification Steps
 
1. Install the operator locally via the command `make create-all` by using this PR and image `docker.io/cmacedo/mobile-security-service-operator:AEROGEAR-9161`
2. Create the namespace for the apps `oc new-project X`
3. Create the app by running `make create-app`
4. Check if the app was created with success ( config map) but it will not be able to save the app in service regards the Oauth impl. 
5. Now, in the same namespace run `make delete-app` it should run but be stuck since it is missing the access to the Service provided by Oauth

## Checklist:

## Progress

- [X] Finished task
- [ ] TODO


1
```diff
@@ -0,0 +1,28 @@
+pick 2d11538 feat(finalizer) : Following the recommendations to work with finalizers
```
austincunningham: don't think this file needs to be committed

camilamacedo86: Agreed on tks to get it.

pkg/controller/mobilesecurityserviceapp/finalizer.go
```diff
@@ -1,14 +1,33 @@
...
-	"fmt"
-	"github.com/aerogear/mobile-security-service-operator/pkg/service"
-	"github.com/go-logr/logr"
 	"context"
+	"fmt"
 	mobilesecurityservicev1alpha1 "github.com/aerogear/mobile-security-service-operator/pkg/apis/mobilesecurityservice/v1alpha1"
+	"github.com/go-logr/logr"
 )
-//updateFinilizer returns error when the app still not deleted in the REST Service
-func (r *ReconcileMobileSecurityServiceApp) updateFinilizer(serviceAPI string,reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityServiceApp) error {
```
damienomurchu:
```suggestion
func (r *ReconcileMobileSecurityServiceApp) updateFinalizer(serviceAPI string,reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityServiceApp) error {
```

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
```diff
@@ -31,4 +31,5 @@ spec:
   #    If the name be not specified then it will be the Mobile Security Service instance name.
   #    The DB will looking for this ConfigMap to get the Env Values variables
   configMapName: "mobile-security-service-config"
-  routeName: "route"
\ No newline at end of file
+  routeName: "route"
+  disableOauth: true
```
davidffrench: This has been added but is not used anywhere

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -154,6 +149,37 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
+				if err := service.DeleteAppFromServiceByRestAPI(serviceAPI,  app.ID, reqLogger); err != nil {
+					reqLogger.Error(err, "App was not delete on the Rest Service", "App.id",  app.ID)
+					return reconcile.Result{}, err
+				}
+				return reconcile.Result{}, nil
+			}
+		}
+
+		//Update finalizer timestamp to allow delete CR
+		if err := r.updateFinalizer(serviceAPI, reqLogger, instance); err != nil {
```
davidffrench: The code here and above was hard to read to figure out what was going on. This seems like it should be the else of the `hasApp(app)` conditional. Although if there is an error from `fetchBindAppRestServiceByAppID` this will also be called, this shouldn't be the case in my opinion.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -154,6 +149,37 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
+		}
+	}
+
+	//Check specs
+	if !hasMandatorySpecs(instance, serviceInstance, reqLogger) {
+		return reconcile.Result{Requeue: true}, nil
+	}
+
+	//Add finalizer for this CR
+	if err := r.addFinalizer(reqLogger, instance); err != nil {
```
davidffrench: Why are we adding a finalizer in the middle of the reconcile loop?

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
camilamacedo86: tks it should not be here :-)

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: Because the finalizer is added when the CR is created. It was recommended to be made in this way.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: @davidffrench I think you did not could check it properly. Following the code and comments.

```go
// Note that it is an if and has a conditional err == nil
if app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger); err == nil {
                          //if has no error will check if app var has APP
			if hasApp(app) {
                                  // if has app.ID then call the DeleteApp
				if err := service.DeleteAppFromServiceByRestAPI(serviceAPI,  app.ID, reqLogger); err != nil {
					reqLogger.Error(err, "App was not delete on the Rest Service", "App.id",  app.ID)
				}
			}
		}
```

I will add a few comments in the code as well.

camilamacedo86: Tks I am updating it.

camilamacedo86:

craicoverflow:
I followed steps 1-4 in the verification steps, but the app does not appear in the UI. Do you know why?

```sh
{"level":"info","ts":1557763898.6658614,"logger":"controller_mobilesecurityserviceapp","msg":"Calling REST API to GET app","Request.Namespace":"mobile-security-service-apps","Request.Name":"mobile-security-service-app1","serviceAPI":"http://route-mobile-security-service-operator.192.168.99.100.nip.io/api","App.appId":"appid1.example.com "}
{"level":"info","ts":1557763898.665887,"logger":"controller_mobilesecurityserviceapp","msg":"URL to get","Request.Namespace":"mobile-security-service-apps","Request.Name":"mobile-security-service-app1","HTTPMethod":"GET","url":"http://route-mobile-security-service-operator.192.168.99.100.nip.io/api/apps?appId=appid1.example.com "}
{"level":"info","ts":1557763898.6927996,"logger":"controller_mobilesecurityserviceapp","msg":"The app was not found in the REST Service API","Request.Namespace":"mobile-security-service-apps","Request.Name":"mobile-security-service-app1","HTTPMethod":"GET","url":"http://route-mobile-security-service-operator.192.168.99.100.nip.io/api/apps?appId=appid1.example.com "}
{"level":"info","ts":1557763898.692882,"logger":"controller_mobilesecurityserviceapp","msg":"Calling REST API to POST app","Request.Namespace":"mobile-security-service-apps","Request.Name":"mobile-security-service-app1","serviceAPI":"http://route-mobile-security-service-operator.192.168.99.100.nip.io/api","App":{"id":"","appId":"appid1.example.com","appName":"example-appname1 "}}
{"level":"info","ts":1557763898.7209384,"logger":"controller_mobilesecurityserviceapp","msg":"Created successfully app object in REST Service API","Request.Namespace":"mobile-security-service-apps","Request.Name":"mobile-security-service-app1","App:":{"id":"","appId":"appid1.example.com","appName":"example-appname1"}}
```

camilamacedo86:
HI @craicoverflow, 

If you follow now with the latest version which is rebased with the master and with the httpcode status and logs improved you will see the error 503 returned from the rest endpoints since the Oauth impl brke the master branch. See https://github.com/aerogear/mobile-security-service-operator/pull/76.

craicoverflow:
> HI @craicoverflow,
> 
> If you follow now with the latest version which is rebased with the master and with the httpcode status and logs improved you will see the error 503 returned from the rest endpoints since the Oauth impl brke the master branch. See #76.

Hi @camilamacedo86 the configmap works fine, but why does the app not show for me on the UI once it is created? It does not get created in the database either.

camilamacedo86:
HI @craicoverflow,

As we chat over the requests has been not authorized by the Oauth which is a bug introduced with this impl. We are working to solve this bug as well :-)

pkg/controller/mobilesecurityserviceapp/helpers.go
```diff
@@ -9,13 +9,20 @@ import (
 )
 
 const SDK  = "-sdk"
+const FINALIZER = "finalizer.mobile-security-service.aerogear.com"
```
craicoverflow: I think it would be a good idea to centrally store values like this in a config object. Currently we are littering these values all over the operator which makes it hard to maintain or change in the future.

The config would be independent from the CR configuration values.

Something like what we have in the [Mobile Security Service](https://github.com/aerogear/mobile-security-service/blob/master/pkg/config/config.go)

This would be out of scope of this PR, but is worth highlighting. What do you think?

camilamacedo86: The reason for it is not a config of the CR and it be a cont is because no have actually a scenario where it would be changed and the other engs do not would like to expose so many config/options.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -154,6 +149,41 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
+			if hasApp(app) {
+				//Send request to delete APP from the Service
+				if err := service.DeleteAppFromServiceByRestAPI(serviceAPI,  app.ID, reqLogger); err != nil {
+					reqLogger.Error(err, "Unable to delete app from Service", "App.ID",  app.ID)
+					return reconcile.Result{}, err
+				}
+				reqLogger.Info("Successfully delete app ...")
+
+				//Update finalizer to allow delete CR
+				if err := r.updateFinalizer(serviceAPI, reqLogger, instance); err != nil {
```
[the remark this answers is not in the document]
camilamacedo86: Is it something that still making sense for you?

pkg/controller/mobilesecurityserviceapp/helpers.go
craicoverflow: Yes as mentioned this config would not be for CRs.

We would not have to expose the options if we do not use environment variables (which we don't have to).

It is very possible that these values could be changed in the future. For example, look at how Mobile Services was renamed to Mobile Developer Services. The same could happen for us, where we rename this from "mobile-security-service" to something new. Then we would have to find every instance of this name throughout the code where if it was in a config it would be easier.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -154,6 +149,41 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
 	//Get the REST Service Endpoint
 	serviceAPI := service.GetServiceAPIURL(route, serviceInstance)
 
+
+	//Check if the APP CR was marked to be deleted
+	isCRMarkedToBeDeleted := instance.GetDeletionTimestamp() != nil
```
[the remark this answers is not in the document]
camilamacedo86: The nomenclature and the process in Kubernetes is marked to delete so I am not sure over this suggestion.

pkg/controller/mobilesecurityserviceapp/helpers.go
camilamacedo86: Hi, @craicoverflow your suggestion is clear. It is not really part of this PR but let's see if we can add it.

pkg/controller/mobilesecurityserviceapp/controller.go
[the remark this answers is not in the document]
craicoverflow: It makes sense, but I still think it should be refactored to reduce the number of nested `if` statements.

pkg/controller/mobilesecurityserviceapp/controller.go
davidffrench: Thanks @camilamacedo86

pkg/controller/mobilesecurityserviceapp/controller.go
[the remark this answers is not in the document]
camilamacedo86: I don't think that it makes sense to me. Because of the if ensuring that the request, for example, will be made just when required and etc .. Are the if that checks in it stage if should or not the code is executed. 

However, feel free to add here your suggestion.

camilamacedo86:
HI @craicoverflow,

Some changes were made based on your comments and a new image build and pushed. Please, feel free to check.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -158,6 +153,38 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
 	//Get the REST Service Endpoint
 	serviceAPI := service.GetServiceAPIURL(route, serviceInstance)
 
+	//Check if the APP CR was marked to be deleted
+	hasDelationMark := instance.GetDeletionTimestamp() != nil
+	if hasDelationMark {
```
davidffrench: @camilamacedo86 Unfortunately the below logic does not cover every failure scenario. Here is a diagram to show what should happen in each case in my opinion. Please let me know if you want to chat about it.

App CR deleteion logic

laurafitzgerald: +1 I'm in the process of testing `make delete-app` against the new no oauth service. I'm seeing some failures with the current logic. I also came across a json.Decoder failure when the response body is `{}` i.e. no app found. It's described here: https://github.com/golang/go/issues/25956, if we continue to use go 1.10 we'll need to cover that as a failure as well with something like
```
if err == io.ErrUnexpectedEOF {
		reqLogger.Info("some informative log")
		return got, nil
	}
```

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -158,6 +153,38 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
 	//Get the REST Service Endpoint
 	serviceAPI := service.GetServiceAPIURL(route, serviceInstance)
 
+	//Check if the APP CR was marked to be deleted
+	hasDelationMark := instance.GetDeletionTimestamp() != nil
```
laurafitzgerald: typo in this variable.

pkg/controller/mobilesecurityserviceapp/finalizer.go
```diff
@@ -3,55 +3,62 @@ package mobilesecurityserviceapp
...
-				if err := service.DeleteAppFromServiceByRestAPI(serviceAPI,  app.ID, reqLogger); err != nil {
-					reqLogger.Error(err, "App was not delete on the Rest Service", "App.id",  app.ID)
-					return err
-				}
-				return nil
-			}
+		reqLogger.Info("Removing Finalizer for the MobileSecurityServiceApp")
+		if app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger);  err != nil || app.ID != "" {
+			err := fmt.Errorf("App was found in the REST Service API")
+			reqLogger.Error(err, "Unable to delete APP", "App.appId", instance.Spec.AppId)
```
laurafitzgerald: I'm not sure if this error log is correct. Based on the logic in the fetch there could be an error returned on fetching the app. If it fails on the fetch then it never tries to do the delete so erroring `Unable to delete APP` seems incorrect to me.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: Hi @davidffrench,

You are correct. I did changes after of the comments raised here and I broke the logic. It is fixed now.

camilamacedo86: Hi @laurafitzgerald,
 
Let's move forward and merge this one to check if the error will be faced. If yes, we can check the error and handle it properly with the upper version. thank you for check it.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: Hi @laurafitzgerald according to the good practices a bool var should be as `isConditional` or `hasConditional`. We should not use the "should" typo as the prefix. 

Also, note the if you read the Kube docs when we use a finalizer it means, at this part of the process, that the CR was marked to be deleted. It dosen't means that should be deleted since it does not mean that all finalizer criteria was performed with success. It is just a flag. 

So, IMHO the best nomenclature here is: `isAppMarkedToBeDeleted`

I hope that it makes sense and that we can agree in move forward.

pkg/controller/mobilesecurityserviceapp/finalizer.go
camilamacedo86: Good spot. Fixed.

pkg/controller/mobilesecurityserviceapp/controller.go
davidffrench: Hi @camilamacedo86, I just checked the code again and there is no handling if there is an error returned from fetching an app from the server.

camilamacedo86: |o| you are correct. I am fixing.

pkg/controller/mobilesecurityserviceapp/controller.go
laurafitzgerald: @camilamacedo86 thanks for the info. In that case `isAppMarkedToBeDeleted` is fine with me.

camilamacedo86: Great !!!!

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -158,6 +153,43 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
+		//If the CR was marked to be deleted before it finalizes the app need to be deleted from the Service
+		//Do request to get the app.ID to delete app
+		app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger);
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		// If the request works with success and the app was found then
+		// Do request to delete it from the service
+		if  app.ID != "" {
```
craicoverflow: There is an extra space between `if` and `app`. On VSCode, the file will automatically format on save. Is there something similar to this for Goland to prevent these things slipping in?

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -158,6 +153,43 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
+			if err := service.DeleteAppFromServiceByRestAPI(serviceAPI,  app.ID, reqLogger); err != nil {
+				reqLogger.Error(err, "Unable to delete app from Service", "App.ID",  app.ID)
+				return reconcile.Result{}, err
+			}
+			reqLogger.Info("Successfully delete app ...")
```
craicoverflow: This could be "succesfully delete**d**" app to indicate the action was a historical event.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: We can add a make command to format the code as it was done in the MSS and do a PR to format all. I do not think we should avoid merge it because of this.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86:
```suggestion
			reqLogger.Info("Successfully deleted app ...")
```

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: Could you please track a task for that?

pkg/controller/mobilesecurityserviceapp/finalizer.go
```diff
@@ -2,56 +2,61 @@ package mobilesecurityserviceapp
...
-	if len(instance.GetFinalizers()) > 0 && instance.GetDeletionTimestamp() != nil {
-		reqLogger.Info("Removing Finalizer for the MobileSecurityServiceApp")
-		if app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger);  err != nil || hasApp(app){
-			if hasApp(app) {
-				err := fmt.Errorf("App was found in the REST Service API")
-				reqLogger.Error(err, "Unable to delete APP", "App.appId", instance.Spec.AppId)
-				return err
-			}
+
+//updateFinalizer returns error when the app still not deleted in the REST Service
```
craicoverflow: The comment still references `updateFinalizer`, whereas the function is now called `addFinalizer`.

pkg/controller/mobilesecurityserviceapp/finalizer.go
```diff
@@ -2,56 +2,61 @@ package mobilesecurityserviceapp
...
-		err := r.client.Update(context.TODO(), instance)
+
+		//Set finalizer string/metadata
+		instance.SetFinalizers([]string{FINALIZER})
+
+		//Update CR
+		err = r.client.Update(context.TODO(), instance)
 		if err != nil {
-			reqLogger.Error(err, "Failed to update finalizer for the MobileSecurityService App")
+			reqLogger.Error(err, "Failed to update MobileSecurityService App CR with  finalizer")
```
craicoverflow: There is an extra space between "with" and "finalizer".

pkg/controller/mobilesecurityserviceapp/finalizer.go
```diff
@@ -2,56 +2,61 @@ package mobilesecurityserviceapp
...
 		}
 	}
 	return nil
 }
 
-//handleFinalizer check if has the finalizer and delete the app
-func (r *ReconcileMobileSecurityServiceApp) handleFinalizer(serviceAPI string,reqLogger logr.Logger, request reconcile.Request) error {
+
+//updateFinalizer returns error when the app still not deleted in the REST Service
+func (r *ReconcileMobileSecurityServiceApp) updateFinalizer(serviceAPI string,reqLogger logr.Logger, request reconcile.Request) error {
```
craicoverflow: Would it be better to call this `removeFinalizer`?

pkg/controller/mobilesecurityserviceapp/finalizer.go
```diff
@@ -2,56 +2,61 @@ package mobilesecurityserviceapp
...
-			if hasApp(app) {
-				if err := service.DeleteAppFromServiceByRestAPI(serviceAPI,  app.ID, reqLogger); err != nil {
-					reqLogger.Error(err, "App was not delete on the Rest Service", "App.id",  app.ID)
-					return err
-				}
-				return nil
-			}
+		reqLogger.Info("Removing Finalizer for the MobileSecurityServiceApp")
+		if app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger);  err != nil || app.ID != "" {
+			reqLogger.Error(err, "Unable to delete APP", "App.appId", instance.Spec.AppId, "app.ID", app.ID)
```
craicoverflow: "APP" does not need to be fully capitalised as it is not an initialism or an acronym.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: Done here : https://issues.jboss.org/browse/AEROGEAR-9213

pkg/controller/mobilesecurityserviceapp/finalizer.go
camilamacedo86: Done 👍

pkg/controller/mobilesecurityserviceapp/finalizer.go
camilamacedo86: Done 👍

pkg/controller/mobilesecurityserviceapp/finalizer.go
camilamacedo86: done

pkg/controller/mobilesecurityserviceapp/finalizer.go
camilamacedo86: Done 👍

pkg/controller/mobilesecurityserviceapp/finalizer.go
camilamacedo86: Done 👍

PR #68, opened 2019-05-09
fix(): Remove hardcoded namespace for the operator (AEROGEAR-9152)

davidffrench:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9152

## What
Where possible, removing hardcoded references to the operator namespace

## Why
Would be unable to deploy the operator in any namespace other than the one hardcoded

## How
Removed where possible from the code

## Verification Steps
1. Config your local env with this PR and the image `docker.io/aerogear/mobile-security-service-operator:0.1.0-AG-9152`
2. Install all by `make create-all`
3. Check the operator is created in the mobile-security-service namespace
5. Check the service and db are provisioned successfully

## Checklist:

- [ ] Code has been tested locally by PR requester
- [ ] Changes have been successfully verified by another team member 

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

PS.: Add images and/or .gifs to illustrate what was changed if this pull request modifies the appearance/output of something presented to the users.

PR #74, opened 2019-05-10
feat(service): Make the service check the http code response and log …

camilamacedo86:
## Motivation
https://github.com/aerogear/mobile-security-service-operator/pull/68

## What
* Make the service validate/check the HTTP code status response
* Promote better logs for the failure situations

## Verification Steps
- It can be checked with this PR and `docker.io/cmacedo/mobile-security-service-operator:FIX-SERVICE` 
- Install the operator `make create-all`
- Run `oc new-project mobile-security-service-apps`
- Then in the valid app namespace try to create an app `make create-app`
- Then in the valid app namespace try to remove the app `make delete-app`

IMPORTANT: It cannot be checked if all works. So, if we the issue be solved in the master branch, https://github.com/aerogear/mobile-security-service-operator/pull/76, and its be rebase we will not able to check what was done here.

Screenshot 2019-05-11 at 00 37 02

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/pkg/service/client.go
+++ b/pkg/service/client.go
@@ -10,13 +10,13 @@ import (
 
 //DeleteAppFromServiceByRestAPI delete the app object in the service
 func DeleteAppFromServiceByRestAPI(serviceAPI string, id string, reqLogger logr.Logger ) error {
-	reqLogger.Info( "Calling REST API to DELETE app", "serviceAPI", serviceAPI, "App.id", id )
+	reqLogger.Info( "Calling REST Service to DELETE app", "serviceAPI", serviceAPI, "App.id", id )
 	//Create the DELETE request
 	url:= serviceAPI + "/apps/" + id
 	req, err := http.NewRequest(http.MethodDelete, url ,nil)
 	req.Header.Set("Content-Type", "application/json")
 	if err != nil {
-		reqLogger.Error(err, "Error when try to create the request", "HTTPMethod", http.MethodDelete, "Request", req, "url", url, "error", err)
+		reqLogger.Error(err, "Unable to create DELETE request", "HTTPMethod", http.MethodDelete, "url", url)
 		return err
 	}
 
@@ -25,19 +25,19 @@ func DeleteAppFromServiceByRestAPI(serviceAPI string, id string, reqLogger logr.
 	response, err := client.Do(req)
 	defer response.Body.Close()
 
-	if err != nil {
-		reqLogger.Error(err, "Error when try to do the request", "HTTPMethod", http.MethodDelete, "Request", req, "url", url, "error", err)
+	if err != nil || 204 != response.StatusCode {
+		reqLogger.Error(err, "HTTP StatusCode not expected", "HTTPMethod", http.MethodDelete, "url", url, "response.StatusCode", response.StatusCode)
 		return err
 	}
 
-	reqLogger.Info("Deleted successfully app object in REST Service API",  "App.Id:", id)
+	reqLogger.Info("Successfully deleted app  ...",  "App.Id:", id)
 	return nil
 }
 
 
 //CreateAppByRestAPI create the app object in the service
 func CreateAppByRestAPI(serviceAPI string, app models.App, reqLogger logr.Logger) error {
-	reqLogger.Info( "Calling REST API to POST app", "serviceAPI", serviceAPI, "App", app )
+	reqLogger.Info( "Calling Service to POST app", "serviceAPI", serviceAPI, "App", app )
 
 	// Create the object and parse for JSON
 	appJSON, err := json.Marshal(app)
@@ -51,20 +51,21 @@ func CreateAppByRestAPI(serviceAPI string, app models.App, reqLogger logr.Logger
 	req, err := http.NewRequest(http.MethodPost, url , strings.NewReader(string(appJSON)))
 	req.Header.Set("Content-Type", "application/json")
 	if err != nil {
-		reqLogger.Error(err, "Error when try to create the request", "HTTPMethod", http.MethodPost, "Request", req, "url", url, "error", err)
+		reqLogger.Error(err, "Unable to create POST request", "HTTPMethod", http.MethodPost, "url", url)
 		return err
 	}
 
 	//Do the request
 	client := &http.Client{}
 	response, err := client.Do(req)
-	if err != nil {
-		reqLogger.Error(err, "Error when try to do the request", "HTTPMethod", http.MethodPost, "Request", req, "Response", response, "url", url, "error", err)
-		return err
-	}
 	defer response.Body.Close()
 
-	reqLogger.Info("Created successfully app object in REST Service API",  "App:", app)
+	if err != nil || 201 != response.StatusCode {
+		reqLogger.Error(err, "HTTP StatusCode not expected", "HTTPMethod", http.MethodPost, "url", url, "response.StatusCode", response.StatusCode )
+		return err
+	}
+
+	reqLogger.Info("Successfully created app  ...",  "App:", app)
 	return nil
 }
 
@@ -76,8 +77,8 @@ func GetAppFromServiceByRestApi(serviceAPI string, appId string, reqLogger logr.
 	// Transform the body request in the version struct
 	got := models.App{}
 
-	if len(appId) < 1 {
-		reqLogger.Info( "Unable to get the AppID", "AppId", appId)
+	if appId == "" {
+		reqLogger.Info( "App without AppId", "App.AppId", appId)
 		return got, nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/service/client.go
```diff
@@ -10,13 +10,13 @@ import (
 
 //DeleteAppFromServiceByRestAPI delete the app object in the service
 func DeleteAppFromServiceByRestAPI(serviceAPI string, id string, reqLogger logr.Logger ) error {
-	reqLogger.Info( "Calling REST API to DELETE app", "serviceAPI", serviceAPI, "App.id", id )
+	reqLogger.Info( "Calling REST Service to DELETE app", "serviceAPI", serviceAPI, "App.id", id )
```
damienomurchu:
```suggestion
	reqLogger.Info( "Calling Service to DELETE app", "serviceAPI", serviceAPI, "App.id", id )
```

Is there a need to specify `REST`? This seems an implementation detail of the client, and if we swap out the current REST implementation in the client for something for something else, like if we changed to using GraphQL it seems we would need to change a lot of logging comments.

damienomurchu:
Changes look good aside from one comment left re use of `REST` in log messages. Haven't verified though.

pkg/service/client.go
camilamacedo86: The logs are important for we are able to troubleshoot it. 
As you can see the changes performed here is more to keep all texts equals and check the HTTP code status for we have sure if it worked or not as throw the errors.

damienomurchu: @camilamacedo86 I'm not sure you understood my point - can you re-read above comment

pkg/service/client.go
```diff
@@ -77,44 +78,42 @@ func GetAppFromServiceByRestApi(serviceAPI string, appId string, reqLogger logr.
...
 	}
 
 	//Do the request
 	client := &http.Client{}
 	response, err := client.Do(req)
 	defer response.Body.Close()
-
 	if err != nil {
-		reqLogger.Error(err, "Error when try to do the request", "HTTPMethod", http.MethodGet, "Request", req, "url", url, "error", err)
+		reqLogger.Error(err, "Unable to do exec GET request", "HTTPMethod", http.MethodGet, "url", url, "response.StatusCode", response.StatusCode)
```
laurafitzgerald: Suggest -> `Unable to execute GET request`

pkg/service/client.go
```diff
@@ -125,7 +124,7 @@ func GetAppFromServiceByRestApi(serviceAPI string, appId string, reqLogger logr.
 
 //deleteAppFromServiceByRestAPI delete the app object in the service
 func UpdateAppNameByRestAPI(serviceAPI string, app models.App, reqLogger logr.Logger) error {
-	reqLogger.Info( "Calling REST API to PATCH app", "serviceAPI", serviceAPI, "App", app )
+	reqLogger.Info( "Calling Rest Service to PATCH app", "serviceAPI", serviceAPI, "App", app )
```
laurafitzgerald: By PATCH here do you mean `update`? Could we use update or POST rather than PATCH?

pkg/service/client.go
```diff
@@ -139,19 +138,19 @@ func UpdateAppNameByRestAPI(serviceAPI string, app models.App, reqLogger logr.Lo
 	req, err := http.NewRequest(http.MethodPatch, url ,strings.NewReader(string(appJSON)))
 	req.Header.Set("Content-Type", "application/json")
 	if err != nil {
-		reqLogger.Error(err, "Error when try to create the request", "HTTPMethod", http.MethodPatch, "Request", req, "url", url, "error", err)
+		reqLogger.Error(err, "Unable to create PATH request to update app name", "HTTPMethod", http.MethodPatch, "url", url)
```
laurafitzgerald: I think `PATH` is a typo here but also see above comment about the use of `PATCH`

pkg/service/client.go
```diff
@@ -139,19 +138,19 @@ func UpdateAppNameByRestAPI(serviceAPI string, app models.App, reqLogger logr.Lo
...
 	response, err := client.Do(req)
 	if err != nil || 204 != response.StatusCode {
-		reqLogger.Error(err, "Error when try to do the request", "HTTPMethod", http.MethodPatch, "Request", req, "url", url, "error", err)
+		reqLogger.Error(err, "HTTP StatusCode not expected", "HTTPMethod", http.MethodPatch, "url", url, "response.StatusCode", response.StatusCode)
 		return err
 	}
 	defer response.Body.Close()
 
-	reqLogger.Info("Deleted successfully app object in REST Service API",  "App:", app)
+	reqLogger.Info("Update successfully app name in Rest Service",  "App", app)
```
laurafitzgerald: Suggestiong -> `Successfully updated app name....`

pkg/service/client.go
```diff
@@ -77,44 +78,42 @@ func GetAppFromServiceByRestApi(serviceAPI string, appId string, reqLogger logr.
...
+	if 200 != response.StatusCode && 204 != response.StatusCode {
+		reqLogger.Error(err, "HTTP StatusCode not expected", "HTTPMethod", http.MethodGet, "url", url, "response.StatusCode", response.StatusCode)
 		return got, nil
 	}
 
 	var obj []models.App
 	err = json.NewDecoder(response.Body).Decode(&obj)
 	if err != nil {
-		reqLogger.Error(err, "Error when try to do decoder the body response", "HTTPMethod", http.MethodGet, "Request", req, "Response", response, "url", url, "error", err)
+		reqLogger.Error(err, "Error when try to do decoder the body response", "HTTPMethod", http.MethodGet, "url", url, "Response.Body", response.Body )
```
laurafitzgerald: decoder -> decode

pkg/service/client.go
```diff
@@ -51,20 +51,21 @@ func CreateAppByRestAPI(serviceAPI string, app models.App, reqLogger logr.Logger
...
+	defer response.Body.Close()
+
+	if err != nil || 201 != response.StatusCode {
+		reqLogger.Error(err, "HTTP StatusCode not expected", "HTTPMethod", http.MethodPost, "url", url, "response.StatusCode", response.StatusCode )
 		return err
 	}
-	defer response.Body.Close()
 
-	reqLogger.Info("Created successfully app object in REST Service API",  "App:", app)
+	reqLogger.Info("Created successfully app object in Rest Service",  "App:", app)
```
laurafitzgerald: Suggestion -> `Successfully created app object....`

pkg/service/client.go
```diff
@@ -25,19 +25,19 @@ func DeleteAppFromServiceByRestAPI(serviceAPI string, id string, reqLogger logr.
...
 
-	if err != nil {
-		reqLogger.Error(err, "Error when try to do the request", "HTTPMethod", http.MethodDelete, "Request", req, "url", url, "error", err)
+	if err != nil || 204 != response.StatusCode {
+		reqLogger.Error(err, "HTTP StatusCode not expected", "HTTPMethod", http.MethodDelete, "url", url, "response.StatusCode", response.StatusCode)
 		return err
 	}
 
-	reqLogger.Info("Deleted successfully app object in REST Service API",  "App.Id:", id)
+	reqLogger.Info("Deleted successfully app object from Rest Service",  "App.Id:", id)
```
laurafitzgerald: Suggested -> `Successfully deleted app....`

pkg/service/client.go
camilamacedo86: PATCH is the type of request.

pkg/service/client.go
camilamacedo86: Done as you suggested.

pkg/service/client.go
camilamacedo86: Done

laurafitzgerald:
@camilamacedo86 regarding the verification of this pr, I've been able to deploy the service successfully. Are there other steps that I should take as I don't see the same logs which are shown in the screenshot in the description.

pkg/service/client.go
camilamacedo86: 👍 done

pkg/service/client.go
camilamacedo86: 👍 done

pkg/service/client.go
camilamacedo86: Done 👍

pkg/service/client.go
camilamacedo86: 👍 done

camilamacedo86:
HI @laurafitzgerald,

The suggested changes in the typos were made. A new image was built and pushed. 
Note that the logs will appear when the service client be in usage which means when you try to create an app and remove an app in the valid namespace.  I will update the first comment.

pkg/service/client.go
```diff
@@ -77,55 +78,53 @@ func GetAppFromServiceByRestApi(serviceAPI string, appId string, reqLogger logr.
...
 		reqLogger.Info("The app was not found in the REST Service API", "HTTPMethod", http.MethodGet, "url", url)
 		return got, nil
 	}
 
-	reqLogger.Info("App found in the Rest Service", "App", obj[0])
+	reqLogger.Info("App found in the Service", "App", obj[0])
 	return obj[0], nil
 }
 
 //deleteAppFromServiceByRestAPI delete the app object in the service
```
davidffrench: Comment needs to be changed

camilamacedo86: Done. Could we move forward now?

camilamacedo86:
@davidffrench all changes requested are done, could you approve it for we are able to move forward?

camilamacedo86:

PR #77, opened 2019-05-12
Refectory controller/reconcile code and fetch instances before updates in order to avoid issues regards the CR be already changed

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9169

## What
- Add export app namespace and remove unnecessary commands
- Refectory controller/reconcile code and fetch instances before updates in order to avoid issues regards the CR be already changed

## Why

**Explanation**
Logs as `Failed to update Status for the Component .... the object has been modified; please apply your changes to the latest version and try again"}` are faced because when the client will update the CR it was already changed. It is like when we go by the OCP UI and try to edit the YAML file of one CR which was changed between the time that the page was rendered and our click on the save button. 
- Try to avoid the above log issues
- Make easier troubleshooting and follow the logs.
- Make easier to keep the project maintained and allow it to be improved. 

## How
- Fetch the latest version of the CR before the updates
- Add comments to explain what is doing in each part
- Improve/Refactory the code 

## Verification Steps
- Install the operator with this PR and image `docker.io/cmacedo/mobile-security-service-operator:AEROGEAR-9169`
- Run `make create-all`
- Check that all was created with success
- In Resources >> other resources: Check that the status was update

Screenshot 2019-05-12 at 20 27 54

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ IMAGE_LATEST_TAG=$(IMAGE_REGISTRY)/$(REGISTRY_ORG)/$(REGISTRY_REPO):latest
 IMAGE_MASTER_TAG=$(IMAGE_REGISTRY)/$(REGISTRY_ORG)/$(REGISTRY_REPO):master
 IMAGE_RELEASE_TAG=$(IMAGE_REGISTRY)/$(REGISTRY_ORG)/$(REGISTRY_REPO):$(CIRCLE_TAG)
 NAMESPACE=mobile-security-service
-APP_NAMESPACE=mobile-security-service-apps
+APP_NAMESPACES=mobile-security-service-apps
 
 # This follows the output format for goreleaser
 BINARY_LINUX_64 = ./dist/linux_amd64/$(BINARY)
@@ -51,8 +51,8 @@ build_linux:
 
 .PHONY: create-app-ns
 create-app-ns:
-	@echo Creating the namespace ${APP_NAMESPACE}:
-	oc new-project ${APP_NAMESPACE}
+	@echo Creating the namespace ${APP_NAMESPACES}:
+	oc new-project ${APP_NAMESPACES}
 
 .PHONY: create-app
 create-app:
@@ -67,7 +67,7 @@ delete-app:
 run-local:
 	@echo Installing the operator in a cluster and run it locally:
 	- export OPERATOR_NAME=mobile-security-service-operator
-	- export APP_NAMESPACES=${APP_NAMESPACE}
+	- export APP_NAMESPACES=${APP_NAMESPACES}
 	- make create-all
 	- operator-sdk up local --namespace=${NAMESPACE}
 
--- a/Makefile
+++ b/Makefile
@@ -62,10 +62,12 @@ create-app:
 delete-app:
 	kubectl delete -f deploy/crds/examples/mobile-security-service_v1alpha1_mobilesecurityserviceapp_cr.yaml
 
+#fixme: The exports in the make file are not working see AEROGEAR-9156. Until it be fixed, run this commands manually.
 .PHONY: run-local
 run-local:
 	@echo Installing the operator in a cluster and run it locally:
 	- export OPERATOR_NAME=mobile-security-service-operator
+	- export APP_NAMESPACES=${APP_NAMESPACE}
 	- make create-all
 	- operator-sdk up local --namespace=${NAMESPACE}
 
@@ -78,8 +80,6 @@ create-all:
 .PHONY: delete-all
 delete-all:
 	@echo Delete Mobile Security Service Operator, Service and namespace "mobile-security-service-operator":
-	- kubectl delete -f deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
-	- kubectl delete -f deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservicedb_cr.yaml
 	- kubectl delete -f deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_crd.yaml
 	- kubectl delete -f deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservicedb_crd.yaml
 	- kubectl delete -f deploy/crds/mobile-security-service_v1alpha1_mobilesecurityserviceapp_crd.yaml
--- a/pkg/controller/mobilesecurityservice/controller.go
+++ b/pkg/controller/mobilesecurityservice/controller.go
@@ -7,7 +7,6 @@ import (
 	"github.com/aerogear/mobile-security-service-operator/pkg/utils"
 	"github.com/go-logr/logr"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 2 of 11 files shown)


pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -146,12 +141,19 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
...
 
-	instance := &mobilesecurityservicev1alpha1.MobileSecurityService{}
-
 	//Fetch the MobileSecurityService instance
+	instance := &mobilesecurityservicev1alpha1.MobileSecurityService{}
 	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
 	if err != nil {
-		return fetch(r, reqLogger, err)
+		instance, err = r.fetchInstance(reqLogger, request)
+		//if errors.IsNotFound(err) {
```
austincunningham: remove commented code

Makefile
```diff
@@ -62,10 +62,12 @@ create-app:
 delete-app:
 	kubectl delete -f deploy/crds/examples/mobile-security-service_v1alpha1_mobilesecurityserviceapp_cr.yaml
 
+#fixme: The exports in the make file are not working see AEROGEAR-9156. Until it be fixed, run this commands manually.
 .PHONY: run-local
 run-local:
 	@echo Installing the operator in a cluster and run it locally:
 	- export OPERATOR_NAME=mobile-security-service-operator
+	- export APP_NAMESPACES=${APP_NAMESPACE}
```
damienomurchu: Where is this new env var (`APP_NAMESPACES`) being used @camilamacedo86 ?

pkg/controller/mobilesecurityservice/controller.go
camilamacedo86: Done .. tks to get it 👍

Makefile
camilamacedo86: It is used to define the namespaces where the app can be installed. It is an ENV VAR of the operator. This command is just for we are able to run the operator locally without a deploy and debug it then because it is required to export the ENV VAR. I think it is explained in the README as well. 

However, note the comment add on top of the command as well. `#fixme: The exports in the make file are not working see AEROGEAR-9156. Until it is fixed, run these commands manually.`

pkg/controller/mobilesecurityservice/status.go
```diff
@@ -3,135 +3,206 @@ package mobilesecurityservice
...
+func (r *ReconcileMobileSecurityService) updateConfigMapStatus(reqLogger logr.Logger, request reconcile.Request) (*corev1.ConfigMap, error) {
 	reqLogger.Info("Updating ConfigMap Status for the MobileSecurityService")
+	// Get the latest version of the CR
+	instance, err := r.fetchInstance(reqLogger, request)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the ConfigMap object
 	configMapStatus, err := r.fetchConfigMap(reqLogger, instance)
```
craicoverflow: Since you are fetching the ConfigMap, could the variable instead just be called `configMap`?

pkg/controller/mobilesecurityservice/status.go
```diff
@@ -3,135 +3,206 @@ package mobilesecurityservice
...
 		return deploymentStatus, err
 	}
-	if !reflect.DeepEqual(deploymentStatus.Name, instance.Status.DeploymentName) {
-		instance.Status.DeploymentName = deploymentStatus.Name
-		err := r.client.Status().Update(context.TODO(), instance)
+
+	// Update the Deployment Name and Status
+	if !reflect.DeepEqual(deploymentStatus.Name, instance.Status.DeploymentName) || !reflect.DeepEqual(deploymentStatus.Status, instance.Status.DeploymentStatus) {
+		// Get the latest version of the CR
+		instance, err := r.fetchInstance(reqLogger, request)
```
davidffrench: I don't believe I understand the exact problem here. Why are we fetching the instance here and on line 94? 

I know the comment reflects that this is an instance of the CR but we should be clear in the variable name

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -193,42 +188,43 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 
 	//Check if Service for the app exist, if not create one
 	if _, err := r.fetchService(reqLogger, instance); err != nil {
```
davidffrench: Is there a scenario here where we would get an error where we should not be creating a new service. The same applies for the deployment and route etc. We should be explicity that if a not found is returned in my opinion. This could be addressed in another PR

pkg/controller/mobilesecurityservice/status.go
craicoverflow: I could be wrong, but I believe it is because there is a potential update that would have occurred on line 65. So the instance would have changed between line 94 and 109.

austincunningham:
needs a rebase

pkg/controller/mobilesecurityservice/status.go
camilamacedo86: Yes. The main goal is to avoid the issues logs when is possible to update a CR that was already changed which indeed will make the process start the reconcile and all checks again.

pkg/controller/mobilesecurityservice/controller.go
camilamacedo86: Yes, an error can be faced.

camilamacedo86:
@craicoverflow do not care over the updateFinilizer it has another PR to be solved.

pkg/controller/mobilesecurityservicedb/controller.go
```diff
@@ -184,36 +187,36 @@ func (r *ReconcileMobileSecurityServiceDB) Reconcile(request reconcile.Request)
...
 
 	//Check if PersistentVolumeClaim for the app exist, if not create one
 	if _, err := r.fetchDBPersistentVolumeClaim(reqLogger, instance); err != nil {
-		return r.create(instance, nil, PVC, reqLogger, err)
+		return r.create(instance, nil, PVC, reqLogger)
 	}
 
 	//Update status for deployment
-	deploymentStatus, err := r.updateDeploymentStatus(reqLogger, instance)
+	deploymentStatus, err := r.updateDeploymentStatus(reqLogger, request)
```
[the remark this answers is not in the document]
camilamacedo86: We could try to do it after.

pkg/controller/mobilesecurityserviceapp/finalizer.go
```diff
@@ -1,14 +1,18 @@
...
 	"fmt"
 	"github.com/aerogear/mobile-security-service-operator/pkg/service"
 	"github.com/go-logr/logr"
-	"context"
-	mobilesecurityservicev1alpha1 "github.com/aerogear/mobile-security-service-operator/pkg/apis/mobilesecurityservice/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 //updateFinilizer returns error when the app still not deleted in the REST Service
-func (r *ReconcileMobileSecurityServiceApp) updateFinilizer(serviceAPI string,reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityServiceApp) error {
+func (r *ReconcileMobileSecurityServiceApp) updateFinilizer(serviceAPI string,reqLogger logr.Logger, request reconcile.Request) error {
```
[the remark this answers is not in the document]
camilamacedo86: We have a PR for the Finalizer stuff do not care over it.

pkg/controller/mobilesecurityservice/status.go
```diff
@@ -3,135 +3,206 @@ package mobilesecurityservice
...
 	if err != nil {
 		reqLogger.Error(err, "Failed to get Deployment for Status", "MobileSecurityService.Namespace", instance.Namespace, "MobileSecurityService.Name", instance.Name)
 		return deploymentStatus, err
 	}
-	if !reflect.DeepEqual(deploymentStatus.Name, instance.Status.DeploymentName) {
-		instance.Status.DeploymentName = deploymentStatus.Name
-		err := r.client.Status().Update(context.TODO(), instance)
+
+	// Update the Deployment Name and Status
+	if !reflect.DeepEqual(deploymentStatus.Name, instance.Status.DeploymentName) || !reflect.DeepEqual(deploymentStatus.Status, instance.Status.DeploymentStatus) {
```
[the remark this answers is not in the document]
camilamacedo86: Let's do it another PR. I made a note about that.

pkg/controller/mobilesecurityservice/status.go
camilamacedo86: We could reuse the same configMap var above to save these values as well.  It is not the goal of this PR but I make a note for we check it in another one.

camilamacedo86:

PR #81, opened 2019-05-15
fix(oauth): allow server access to the operator

laurafitzgerald:
## Motivation
[JIRA Comment](https://issues.jboss.org/browse/AEROGEAR-8987?focusedCommentId=13732533&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13732533)

Please note there are some changes here to the creation and deletion logic. These can be removed based on the other changes being made by @camilamacedo86 in another pr.

## What
An additional Service is added by the operator which is not fronted by a Route to be used by the operator in order to interact with the service as required by the consumption pr.

## Verification Steps

1. Using this image: `quay.io/laurafitzgerald/mobile-security-service-operator:0.1.0-dev-ag8987-2` 
2. Run `make create-all`
3. Verify access to the server and additionally that the make delete and update of apps is working correctly.
a) a new Service is created titled `security-service-server` which can be seen in the console
b) Operator updates the app in the service as required by the state of the app cr
   i) create app
   `oc new-project mobile-security-service-apps`
   `make create-app`
   Verify the app has been created in the UI
   ii) update app
   Update the name in the app cr
   Verify the app name is updated in the UI
   iii) delete app
  `make delete-app`
   Verify the app has been deleted from the UI

## Checklist:

### DIFF
--- a/pkg/controller/mobilesecurityservice/controller.go
+++ b/pkg/controller/mobilesecurityservice/controller.go
@@ -18,11 +18,11 @@ import (
 )
 
 const (
-	CONFIGMAP     = "ConfigMap"
-	DEEPLOYMENT   = "Deployment"
-	SERVICE       = "Service"
-	ROUTE         = "Route"
-	SERVERSERVICE = "Server Service"
+	CONFIGMAP           = "ConfigMap"
+	DEEPLOYMENT         = "Deployment"
+	PROXY_SERVICE       = "Proxy Service"
+	ROUTE               = "Route"
+	APPLICATION_SERVICE = "Application Service"
 )
 
 var log = logf.Log.WithName("controller_mobilesecurityservice")
@@ -115,22 +115,22 @@ func (r *ReconcileMobileSecurityService) create(instance *mobilesecurityservicev
 	return reconcile.Result{Requeue: true}, nil
 }
 
-// buildFactory will return the resource according to the kind defined
-func (r *ReconcileMobileSecurityService) buildFactory(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService, kind string) (runtime.Object, error) {
-	reqLogger.Info("Check "+kind, "into the namespace", instance.Namespace)
-	switch kind {
+// buildFactory will return the resource according to the resource defined
+func (r *ReconcileMobileSecurityService) buildFactory(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService, resource string) (runtime.Object, error) {
+	reqLogger.Info("Check "+resource, "into the namespace", instance.Namespace)
+	switch resource {
 	case CONFIGMAP:
 		return r.buildConfigMap(instance), nil
 	case DEEPLOYMENT:
 		return r.buildDeployment(instance), nil
-	case SERVICE:
-		return r.buildService(instance), nil
-	case SERVERSERVICE:
-		return r.buildServerService(instance), nil
+	case PROXY_SERVICE:
+		return r.buildProxyService(instance), nil
+	case APPLICATION_SERVICE:
+		return r.buildApplicationService(instance), nil
 	case ROUTE:
 		return r.buildRoute(instance), nil
 	default:
-		msg := "Failed to recognize type of object" + kind + " into the Namespace " + instance.Namespace
+		msg := "Failed to recognize type of object" + resource + " into the Namespace " + instance.Namespace
 		panic(msg)
 	}
 }
@@ -190,13 +190,13 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 	}
 
 	//Check if Service for the app exist, if not create one
-	if _, err := r.fetchService(reqLogger, instance); err != nil {
-		return r.create(instance, reqLogger, SERVICE)
+	if _, err := r.fetchService(reqLogger, instance, utils.PROXY_SERVICE_INSTANCE_NAME); err != nil {
+		return r.create(instance, reqLogger, PROXY_SERVICE)
 	}
 
 	//Check if Service for the app exist, if not create one
-	if _, err := r.fetchServerService(reqLogger, instance); err != nil {
-		return r.create(instance, reqLogger, SERVERSERVICE)
+	if _, err := r.fetchService(reqLogger, instance, utils.APPLICATION_SERVICE_INSTANCE_NAME); err != nil {
+		return r.create(instance, reqLogger, APPLICATION_SERVICE)
 	}
 
 	//Check if Route for the Service exist, if not create one
@@ -216,14 +216,13 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 		return reconcile.Result{}, err
 	}
 
-	//Update status for Service
-	serviceStatus, err := r.updateServiceStatus(reqLogger, request)
-
+	//Update status for Proxy Service
+	proxyServiceStatus, err := r.updateServiceStatus(reqLogger, request, utils.PROXY_SERVICE_INSTANCE_NAME)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
-	serverServiceStatus, err := r.updateServiceStatus(reqLogger, request)
+	applicationServiceStatus, err := r.updateServiceStatus(reqLogger, request, utils.APPLICATION_SERVICE_INSTANCE_NAME)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
@@ -235,7 +234,7 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 	}
 
 	//Update status for App
-	if err := r.updateStatus(reqLogger, configMapStatus, deploymentStatus, serviceStatus, serverServiceStatus, routeStatus, request); err != nil {
+	if err := r.updateStatus(reqLogger, configMapStatus, deploymentStatus, proxyServiceStatus, applicationServiceStatus, routeStatus, request); err != nil {
 
 		return reconcile.Result{}, err
 	}
--- a/pkg/controller/mobilesecurityservice/controller.go
+++ b/pkg/controller/mobilesecurityservice/controller.go
@@ -18,10 +18,11 @@ import (
 )
 
 const (
-	CONFIGMAP   = "ConfigMap"
-	DEEPLOYMENT = "Deployment"
-	SERVICE     = "Service"
-	ROUTE       = "Route"
+	CONFIGMAP     = "ConfigMap"
+	DEEPLOYMENT   = "Deployment"
+	SERVICE       = "Service"
+	ROUTE         = "Route"
+	SERVERSERVICE = "Server Service"
 )
 
 var log = logf.Log.WithName("controller_mobilesecurityservice")
@@ -124,6 +125,8 @@ func (r *ReconcileMobileSecurityService) buildFactory(reqLogger logr.Logger, ins
 		return r.buildDeployment(instance), nil
 	case SERVICE:
 		return r.buildService(instance), nil
+	case SERVERSERVICE:
+		return r.buildServerService(instance), nil
 	case ROUTE:
 		return r.buildRoute(instance), nil
 	default:
@@ -191,6 +194,11 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 		return r.create(instance, reqLogger, SERVICE)
 	}
 
+	//Check if Service for the app exist, if not create one
+	if _, err := r.fetchServerService(reqLogger, instance); err != nil {
+		return r.create(instance, reqLogger, SERVERSERVICE)
+	}
+
 	//Check if Route for the Service exist, if not create one
 	if _, err := r.fetchRoute(reqLogger, instance); err != nil {
 		return r.create(instance, reqLogger, ROUTE)
@@ -210,6 +218,12 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 
 	//Update status for Service
 	serviceStatus, err := r.updateServiceStatus(reqLogger, request)
+
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	serverServiceStatus, err := r.updateServiceStatus(reqLogger, request)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
@@ -221,7 +235,8 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 	}
 
 	//Update status for App
-	if err := r.updateStatus(reqLogger, configMapStatus, deploymentStatus, serviceStatus, routeStatus, request); err != nil {
+	if err := r.updateStatus(reqLogger, configMapStatus, deploymentStatus, serviceStatus, serverServiceStatus, routeStatus, request); err != nil {
+
 		return reconcile.Result{}, err
 	}
 
--- a/pkg/controller/mobilesecurityservice/fetches.go
+++ b/pkg/controller/mobilesecurityservice/fetches.go
@@ -30,19 +30,11 @@ func (r *ReconcileMobileSecurityService) fetchRoute(reqLogger logr.Logger, insta
 	return route, err
 }
 
-//fetchService returns the service resource created for this instance
-func (r *ReconcileMobileSecurityService) fetchService(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService) (*corev1.Service, error) {
+//fetchServerService returns the application service resource created for this instance
+func (r *ReconcileMobileSecurityService) fetchService(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService, name string) (*corev1.Service, error) {
 	reqLogger.Info("Checking if the service already exists")
 	service := &corev1.Service{}
-	err := r.client.Get(context.TODO(), types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}, service)
-	return service, err
-}
-
-//fetchServerService returns the service resource created for this instance
-func (r *ReconcileMobileSecurityService) fetchServerService(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService) (*corev1.Service, error) {
-	reqLogger.Info("Checking if the service already exists")
-	service := &corev1.Service{}
-	err := r.client.Get(context.TODO(), types.NamespacedName{Name: utils.SERVER_SERVICE_INSTANCE_NAME, Namespace: instance.Namespace}, service)
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: instance.Namespace}, service)
 	return service, err
 }
 
--- a/pkg/controller/mobilesecurityservice/fetches.go
+++ b/pkg/controller/mobilesecurityservice/fetches.go
@@ -2,6 +2,7 @@ package mobilesecurityservice
 
 import (
 	"context"
+
 	mobilesecurityservicev1alpha1 "github.com/aerogear/mobile-security-service-operator/pkg/apis/mobilesecurityservice/v1alpha1"
 	"github.com/aerogear/mobile-security-service-operator/pkg/utils"
 	"github.com/go-logr/logr"
@@ -14,13 +15,12 @@ import (
 
 // Request object not found, could have been deleted after reconcile request.
 // Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
-func (r *ReconcileMobileSecurityService) fetchInstance( reqLogger logr.Logger, request reconcile.Request) (*mobilesecurityservicev1alpha1.MobileSecurityService, error) {
+func (r *ReconcileMobileSecurityService) fetchInstance(reqLogger logr.Logger, request reconcile.Request) (*mobilesecurityservicev1alpha1.MobileSecurityService, error) {
 	instance := &mobilesecurityservicev1alpha1.MobileSecurityService{}
 	//Fetch the MobileSecurityService instance
 	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
 	return instance, err
 }
-
 
 //fetchRoute returns the Route resource created for this instance
 func (r *ReconcileMobileSecurityService) fetchRoute(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService) (*routev1.Route, error) {
@@ -35,6 +35,14 @@ func (r *ReconcileMobileSecurityService) fetchService(reqLogger logr.Logger, ins
 	reqLogger.Info("Checking if the service already exists")
 	service := &corev1.Service{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}, service)
+	return service, err
+}
+
+//fetchServerService returns the service resource created for this instance
+func (r *ReconcileMobileSecurityService) fetchServerService(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService) (*corev1.Service, error) {
+	reqLogger.Info("Checking if the service already exists")
+	service := &corev1.Service{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: utils.SERVER_SERVICE_INSTANCE_NAME, Namespace: instance.Namespace}, service)
 	return service, err
 }
 
@@ -53,5 +61,3 @@ func (r *ReconcileMobileSecurityService) fetchConfigMap(reqLogger logr.Logger, i
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: utils.GetConfigMapName(instance), Namespace: instance.Namespace}, configMap)
 	return configMap, err
 }
-
-
--- a/pkg/controller/mobilesecurityservice/services.go
+++ b/pkg/controller/mobilesecurityservice/services.go
@@ -2,6 +2,7 @@ package mobilesecurityservice
 
 import (
 	mobilesecurityservicev1alpha1 "github.com/aerogear/mobile-security-service-operator/pkg/apis/mobilesecurityservice/v1alpha1"
+	"github.com/aerogear/mobile-security-service-operator/pkg/utils"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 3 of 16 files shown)


pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -124,6 +125,8 @@ func (r *ReconcileMobileSecurityService) buildFactory(reqLogger logr.Logger, ins
 		return r.buildDeployment(instance), nil
 	case SERVICE:
 		return r.buildService(instance), nil
+	case SERVERSERVICE:
```
camilamacedo86: What is the ServiceServer>? Is The Service to expose the deployment of MSS service?

pkg/utils/utils.go
```diff
@@ -15,6 +15,7 @@ import (
 // The namespaces should be informed split by ";".
 const APP_NAMESPACE_ENV_VAR = "APP_NAMESPACES"
 const SERVICE_INSTANCE_NAME = "mobile-security-service"
+const SERVER_SERVICE_INSTANCE_NAME = "mobile-security-server"
```
camilamacedo86:
```suggestion
const BYPASS_SERVICE_INSTANCE_NAME = "mobile-security-service-baypass"
```

pkg/service/utils.go
```diff
@@ -1,20 +1,15 @@
...
-//Return the router host URL
-func GetServiceURL(route *routev1.Route, serviceInstance *mobilesecurityservicev1alpha1.MobileSecurityService) string{
-	return serviceInstance.Spec.ClusterProtocol + "://" + route.Status.Ingress[0].Host
-}
-
 //Return REST Service API
-func GetServiceAPIURL(route *routev1.Route, serviceInstance *mobilesecurityservicev1alpha1.MobileSecurityService) string{
-	return GetServiceURL(route,serviceInstance) + ENDPOINT_API
+func GetServiceAPIURL(serviceInstance *mobilesecurityservicev1alpha1.MobileSecurityService) string {
+	return serviceInstance.Spec.ClusterProtocol + "://" + utils.SERVER_SERVICE_INSTANCE_NAME + ":" + fmt.Sprint(serviceInstance.Spec.Port) + ENDPOINT_API
```
camilamacedo86: It is used in the SDKConfigMap. So, could you keep on impl to attend the SDKConfigMap?

pkg/service/client.go
```diff
@@ -108,34 +110,40 @@ func GetAppFromServiceByRestApi(serviceAPI string, appId string, reqLogger logr.
 
 	var obj []models.App
 	err = json.NewDecoder(response.Body).Decode(&obj)
-	if err != nil {
-		reqLogger.Error(err, "Error when try to do decode the body response", "HTTPMethod", http.MethodGet, "url", url, "Response.Body", response.Body )
-		return got, err
+
+	if 204 == response.StatusCode {
```
camilamacedo86: We need to check the decode error before the httpCode status.

pkg/service/client.go
```diff
@@ -95,7 +98,6 @@ func GetAppFromServiceByRestApi(serviceAPI string, appId string, reqLogger logr.
 	//Do the request
 	client := &http.Client{}
 	response, err := client.Do(req)
-	defer response.Body.Close()
```
camilamacedo86: Should we really remove it?

pkg/controller/mobilesecurityserviceapp/helpers.go
```diff
@@ -36,8 +37,12 @@ func getSDKConfigMapName(m *mobilesecurityservicev1alpha1.MobileSecurityServiceA
 }
 
 //hasApp return true when APP has ID which is just created by the REST Service API
-func hasApp(app models.App) bool {
-	return len(app.ID) > 0
+func hasApp(app *models.App) bool {
```
camilamacedo86: It was removed. @craicoverflow suggested that was easier to understand seen thee check app.ID == ""for example and I agreed which make this func unnecessary.

pkg/controller/mobilesecurityserviceapp/finalizer.go
```diff
@@ -35,23 +40,30 @@ func (r *ReconcileMobileSecurityServiceApp) updateFinilizer(serviceAPI string,re
 }
 
 //handleFinalizer check if has the finalizer and delete the app
-func (r *ReconcileMobileSecurityServiceApp) handleFinalizer(serviceAPI string,reqLogger logr.Logger, request reconcile.Request) error {
+func (r *ReconcileMobileSecurityServiceApp) handleFinalizer(serviceAPI string, reqLogger logr.Logger, request reconcile.Request) error {
```
camilamacedo86: I think we should apply the PR created to improve and make it works according to the recommendations.
Wdyt over merge https://github.com/aerogear/mobile-security-service-operator/pull/73 then you rebase with the master and check if you would like to perform more changes?

pkg/controller/mobilesecurityservice/status.go
```diff
@@ -22,12 +23,12 @@ func (r *ReconcileMobileSecurityService) updateStatus(reqLogger logr.Logger, con
 	}
 
 	//Check if all required objects are created
-	if len(configMapStatus.UID) < 1 && len(deploymentStatus.UID) < 1 && len(serviceStatus.UID) < 1 && len(routeStatus.Name) < 1 {
+	if len(configMapStatus.UID) < 1 && len(deploymentStatus.UID) < 1 && len(serviceStatus.UID) < 1 && len(serverServiceStatus.UID) < 1 && len(routeStatus.Name) < 1 {
```
camilamacedo86:
```suggestion
	if len(configMapStatus.UID) < 1 && len(deploymentStatus.UID) < 1 && len(serviceStatus.UID) < 1 && len(bypassServiceStatus.UID) < 1 && len(routeStatus.Name) < 1 {
```

pkg/controller/mobilesecurityservice/fetches.go
```diff
@@ -38,6 +38,14 @@ func (r *ReconcileMobileSecurityService) fetchService(reqLogger logr.Logger, ins
 	return service, err
 }
 
+//fetchServerService returns the service resource created for this instance
+func (r *ReconcileMobileSecurityService) fetchServerService(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService) (*corev1.Service, error) {
+	reqLogger.Info("Checking if the service already exists")
```
camilamacedo86:
```suggestion
	reqLogger.Info("Checking if the bypass service already exists")
```

pkg/controller/mobilesecurityservice/fetches.go
```diff
@@ -38,6 +38,14 @@ func (r *ReconcileMobileSecurityService) fetchService(reqLogger logr.Logger, ins
 	return service, err
 }
 
+//fetchServerService returns the service resource created for this instance
+func (r *ReconcileMobileSecurityService) fetchServerService(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService) (*corev1.Service, error) {
```
camilamacedo86:
```suggestion
func (r *ReconcileMobileSecurityService) fetchBypassrService(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService) (*corev1.Service, error) {
```

pkg/controller/mobilesecurityservice/services.go
```diff
@@ -41,3 +42,33 @@ func (r *ReconcileMobileSecurityService) buildService(m *mobilesecurityservicev1
 	controllerutil.SetControllerReference(m, ser, r.scheme)
 	return ser
 }
+
+func (r *ReconcileMobileSecurityService) buildServerService(m *mobilesecurityservicev1alpha1.MobileSecurityService) *corev1.Service {
```
camilamacedo86:
```suggestion
func (r *ReconcileMobileSecurityService) buildBypassService(m *mobilesecurityservicev1alpha1.MobileSecurityService) *corev1.Service {
```

pkg/controller/mobilesecurityservice/services.go
```diff
@@ -41,3 +42,33 @@ func (r *ReconcileMobileSecurityService) buildService(m *mobilesecurityservicev1
...
+	ls := getAppLabels(m.Name)
+	ser := &corev1.Service{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Service",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      utils.SERVER_SERVICE_INSTANCE_NAME,
```
camilamacedo86:
```suggestion
			Name:      utils.BYPASS_SERVICE_INSTANCE_NAME,
```

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -191,6 +194,11 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 		return r.create(instance, reqLogger, SERVICE)
 	}
 
+	//Check if Service for the app exist, if not create one
+	if _, err := r.fetchServerService(reqLogger, instance); err != nil {
```
camilamacedo86:
```suggestion
	if _, err := r.fetchBypassService(reqLogger, instance); err != nil {
```

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -18,10 +18,11 @@ import (
...
 const (
-	CONFIGMAP   = "ConfigMap"
-	DEEPLOYMENT = "Deployment"
-	SERVICE     = "Service"
-	ROUTE       = "Route"
+	CONFIGMAP     = "ConfigMap"
+	DEEPLOYMENT   = "Deployment"
+	SERVICE       = "Service"
+	ROUTE         = "Route"
+	SERVERSERVICE = "Server Service"
```
camilamacedo86:
```suggestion
	BYPASS_SERVICE = "Bypass Service"
```

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -124,6 +125,8 @@ func (r *ReconcileMobileSecurityService) buildFactory(reqLogger logr.Logger, ins
 		return r.buildDeployment(instance), nil
 	case SERVICE:
 		return r.buildService(instance), nil
+	case SERVERSERVICE:
+		return r.buildServerService(instance), nil
```
camilamacedo86:
```suggestion
		return r.buildBypassService(instance), nil
```

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -210,7 +218,12 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 	}
 
 	//Update status for Service
-	serviceStatus, err := r.updateServiceStatus(reqLogger,request)
+	serviceStatus, err := r.updateServiceStatus(reqLogger, request)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	serverServiceStatus, err := r.updateServiceStatus(reqLogger, request)
```
camilamacedo86: @laurafitzgerald,
We cannot use the same func to update both. It means that the second will subscribe to the ServiceStatus of the first one.

camilamacedo86:
Hi @laurafitzgerald fantastic work 👍 
It shows fine, I'd like to do a few suggestions.
1. Replace the name ServerService by BypassService in order to make clear what it is 
2. I think we should merge the PR https://github.com/aerogear/mobile-security-service-operator/pull/73 which has the impl to make the finalizer follow its recommendations and then if you think that still a need to perform changes or improvements please feel free. 
3. The GetApiURL is used in the SDKConfigMap. Did you check the values on it? I think we need to keep an impl for it still return the data from the route as it was done before. 
4. Shows that you are using the fun updateServiceStatus for the new service and we need to have one specific for each

pkg/controller/mobilesecurityservice/controller.go
laurafitzgerald: Yes exactly. :) The naming might not be the best for some of the new functions objects. Happy to discuss any refactoring we need there. e.g. we might want to rename SERVICE now to APPLICATIONSERVICE or similar. Now that the changes are made I will have a little dig in other operators to see how that is managed. I imagine there are other operators which deploy more than one of a particular OpenShift resource.

camilamacedo86: But what is the goal of this service? Could we use a name that makes clear that it is just to bypass? Because the APP service in my POV is the service that exposes the APP Deployment which is not this one. It is more a workarround to allow the APP CR/Operator be able to do the requests by passing the oauth.Wdyt?

grdryn: My suggestion would be to rename both this and the pre-existing `SERVICE` variable above it also. Maybe `MSS_SERVICE` here and `PROXY_SERVICE` for the other one?

It's possible that some other things about this function need to change, since we're switching on a string called `kind`, but both of these are really the same k8s `Kind`, so that's either not a correct name or more logic is needed.

pkg/service/client.go
laurafitzgerald: @camilamacedo86 it's not removed, it is moved below some other logic.

pkg/controller/mobilesecurityservice/fetches.go
```diff
@@ -38,6 +38,14 @@ func (r *ReconcileMobileSecurityService) fetchService(reqLogger logr.Logger, ins
...
+	reqLogger.Info("Checking if the service already exists")
+	service := &corev1.Service{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: utils.SERVER_SERVICE_INSTANCE_NAME, Namespace: instance.Namespace}, service)
+	return service, err
+}
```
grdryn: Suggestion: change the existing `fetchService` function to accept `name` and `namespace` string parameters rather than the `instance` parameter (which is only used to get those strings). Then you can use the same function for both.

pkg/controller/mobilesecurityserviceapp/helpers.go
laurafitzgerald: 👍 is that merged? I probably need to rebase now based on your changes.

pkg/controller/mobilesecurityserviceapp/finalizer.go
laurafitzgerald: @camilamacedo86 yes 100% agree to merge that. Some of the changes in this pr will then be obsolete which I noted in the description.

pkg/controller/mobilesecurityservice/services.go
```diff
@@ -41,3 +42,33 @@ func (r *ReconcileMobileSecurityService) buildService(m *mobilesecurityservicev1
 	controllerutil.SetControllerReference(m, ser, r.scheme)
 	return ser
 }
+
+func (r *ReconcileMobileSecurityService) buildServerService(m *mobilesecurityservicev1alpha1.MobileSecurityService) *corev1.Service {
+	ls := getAppLabels(m.Name)
+	ser := &corev1.Service{
+		TypeMeta: metav1.TypeMeta{
```
grdryn: I think this `TypeMeta` stuff is implicit (you are creating a `corev1.Service` after all). At least I've never bothered setting it and everything works fine. It's fine to leave it there though.

pkg/controller/mobilesecurityservice/services.go
```diff
@@ -41,3 +42,33 @@ func (r *ReconcileMobileSecurityService) buildService(m *mobilesecurityservicev1
...
+			Kind:       "Service",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      utils.SERVER_SERVICE_INSTANCE_NAME,
+			Namespace: m.Namespace,
+			Labels:    ls,
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: ls,
+			Type:     corev1.ServiceTypeClusterIP,
```
grdryn: again, optional, setting to a default

pkg/controller/mobilesecurityservice/services.go
```diff
@@ -41,3 +42,33 @@ func (r *ReconcileMobileSecurityService) buildService(m *mobilesecurityservicev1
...
+			Ports: []corev1.ServicePort{
+				{
+					Port:     m.Spec.Port,
+					Protocol: "TCP",
+					Name:     "server",
+				},
+			},
+			SessionAffinity: "None",
```
grdryn: again, optional, setting to a default

pkg/controller/mobilesecurityservice/services.go
```diff
@@ -41,3 +42,33 @@ func (r *ReconcileMobileSecurityService) buildService(m *mobilesecurityservicev1
...
+			Namespace: m.Namespace,
+			Labels:    ls,
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: ls,
+			Type:     corev1.ServiceTypeClusterIP,
+			Ports: []corev1.ServicePort{
+				{
+					Port:     m.Spec.Port,
+					Protocol: "TCP",
```
grdryn: again, optional, setting to a default

pkg/controller/mobilesecurityserviceapp/finalizer.go
```diff
@@ -3,29 +3,34 @@ package mobilesecurityserviceapp
...
-		if app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger);  err != nil || hasApp(app){
-			if hasApp(app) {
-				err := fmt.Errorf("App was found in the REST Service API")
-				reqLogger.Error(err, "Unable to delete APP", "App.appId", instance.Spec.AppId)
-				return err
-			}
+		app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger)
+		if err != nil {
+			reqLogger.Info("updated finalizer - Request for app by id failed")
+			//TODO we need to handle this error
```
grdryn: Does this TODO need to be TODOne here?

pkg/controller/mobilesecurityserviceapp/finalizer.go
```diff
@@ -35,23 +40,30 @@ func (r *ReconcileMobileSecurityServiceApp) updateFinilizer(serviceAPI string,re
...
-				if err := service.DeleteAppFromServiceByRestAPI(serviceAPI,  app.ID, reqLogger); err != nil {
-					reqLogger.Error(err, "App was not delete on the Rest Service", "App.id",  app.ID)
-					return err
-				}
-				return nil
+		app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger)
+
+		if err != nil {
+			reqLogger.Info("handle finalizer - Request for app by id failed")
+			//TODO we need to handle this error
```
grdryn: Same TODO as above

laurafitzgerald:
> Hi @laurafitzgerald fantastic work 👍
> It shows fine, I'd like to do a few suggestions.
> 
> 1. Replace the name ServerService by BypassService in order to make clear what it is
> 2. I think we should merge the PR #73  which has the impl to make the finalizer follow its recommendations and then if you think that still a need to perform changes or improvements please feel free. Wdyt?
> 3. The GetApiURL is used in the SDKConfigMap. Did you check the values on it? I think we need to keep an impl for it still return the data from the route as it was done before.
> 4. Shows that you are using the fun updateServiceStatus for the new service and we need to have one specific for each
> 
> Please, let me know if I can do anything to help with.

Hi @camilamacedo86 thanks for the review.

1. +1 on renaming, I've skipped over a lot of the review items as these will change once I do the rename. There seems to be a couple of recommendations around this so will review and chose one.
2. +1 do you need me to do anything in order to get that merged? I believe that some of the changes in this pr can be dropped based on that. 
3. I've added it as a constant to the utils pkg so that it's available to both mobilesecurityserviceapp and mobilesecurityservice. I will investigate storing it the config map instead. 
4. If it's passing the instance does that not mean that it will check each service seperately. 
As @grdryn mentioned there could be some refactoring done here. Before I rebased I was passing in the name of the service and calling the generic steps and passing the name down. Will have a look to see how we can improve this.

I think the best way to proceed is to get your pr merged so I can do a rebase and understand what changes here can be dropped.

pkg/controller/mobilesecurityservice/services.go
camilamacedo86: It is a good spot. I did not check if it is implicit.
We definitely should check it and if yes, go across the project removing it since will be an unnecessary code and probably not the best way.

camilamacedo86: Hi @grdryn excellent spot !!! Very good that you catch it. 
The change is here: https://github.com/aerogear/mobile-security-service-operator/pull/84

camilamacedo86:
@laurafitzgerald 

Regards item 4. It is not a check. The impl will update the field ServiceStatus in the CR with the Status of the object returned by the IP. In this way, is required to create a new Status for the Proxy Service and then do an impl to update it. Please, let me know if it makes sense for you and if not let's chat by BJ.

davidffrench:
@laurafitzgerald Once https://github.com/aerogear/mobile-security-service-operator/pull/73 is merged into master and this is rebased, please let me know and I will verify this PR.

pkg/controller/mobilesecurityservice/controller.go
laurafitzgerald: I'm in agreement that kind might not be fully appropriate here. Depending on the application it may be required to deploy >1 of some resources. I think resource might be more appropriate here and is understood in k8s/openshift terminology - e.g. `Other Resources` option in the menu in openshift.

camilamacedo86: I think we can try to improve as recommended by @grdryn. However, IMHO just rename and keep one func for each one then after we can check the best approach to improve and etc .. in another PR

pkg/controller/mobilesecurityservice/status.go
```diff
@@ -22,20 +23,15 @@ func (r *ReconcileMobileSecurityService) updateStatus(reqLogger logr.Logger, con
...
 		err := fmt.Errorf("Failed to get OK Status for MobileSecurityService")
 		reqLogger.Error(err, "One of the resources are not created", "MobileSecurityService.Namespace", instance.Namespace, "MobileSecurityService.Name", instance.Name)
 		return err
 	}
-	status:= "OK"
+	status := "OK"
 
 	// Update CR with the AppStatus == OK
 	if !reflect.DeepEqual(status, instance.Status.AppStatus) {
-		// Get the latest version of the CR
```
laurafitzgerald: It looks like I'm removing this but the instance is already fetched earlier on in the block of code.

pkg/controller/mobilesecurityservice/status.go
```diff
@@ -127,27 +117,22 @@ func (r *ReconcileMobileSecurityService) updateDeploymentStatus(reqLogger logr.L
...
+func (r *ReconcileMobileSecurityService) updateServiceStatus(reqLogger logr.Logger, request reconcile.Request, name string) (*corev1.Service, error) {
+	reqLogger.Info("Updating ", name, " Service Status for the MobileSecurityService")
 	// Get the latest version of the CR
 	instance, err := r.fetchInstance(reqLogger, request)
 	if err != nil {
 		return nil, err
 	}
 	// Get the Service Object
-	serviceStatus, err := r.fetchService(reqLogger, instance)
+	serviceStatus, err := r.fetchService(reqLogger, instance, name)
```
laurafitzgerald: passsing the name of the service here so this method can be used more generically. It can probably be improved further as suggested by @grdryn but wanting to push the current implementation. Will look into this later.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -136,43 +135,38 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
 	reqLogger.Info("Valid namespace for MobileSecurityServiceApp", "Namespace", request.Namespace)
 
 	reqLogger.Info("Checking for service instance ...")
-	serviceInstance := &mobilesecurityservicev1alpha1.MobileSecurityService{}
+	mssInstance := &mobilesecurityservicev1alpha1.MobileSecurityService{}
```
laurafitzgerald: renaming the variable here as the instance gotten is an instance of the MobileSecurityService{} object. As we have two k8s service objects, I'm renaming to avoid confusion.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -136,43 +135,38 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: utils.SERVICE_INSTANCE_NAME, Namespace: operatorNamespace}, serviceInstance); err != nil {
+
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "mobile-security-service", Namespace: operatorNamespace}, mssInstance); err != nil {
 		// Return and don't create
-		reqLogger.Info("Mobile Security Service instance resource not found. Ignoring since object must be deleted")
+		reqLogger.Info("Mobile Security Service instance resource not found. Mobile Security Service Application is required to create the application")
 		return reconcile.Result{}, nil
 	}
 
-	reqLogger.Info("Checking if the route already exists ...")
```
laurafitzgerald: The app controller doesn't use the route because it's using the internal service.

pkg/service/client.go
```diff
@@ -59,90 +61,86 @@ func CreateAppByRestAPI(serviceAPI string, app models.App, reqLogger logr.Logger
...
-	reqLogger.Info("Successfully created app  ...",  "App:", app)
+	defer response.Body.Close()
+
+	reqLogger.Info("Successfully created app  ...", "App:", app)
 	return nil
 }
 
 //GetAppFromServiceByRestApi returns the app object from the service
-func GetAppFromServiceByRestApi(serviceAPI string, appId string, reqLogger logr.Logger) (models.App, error) {
-	reqLogger.Info( "Calling REST API to GET app", "serviceAPI", serviceAPI, "App.appId", appId )
```
laurafitzgerald: Moving this log down as it gets logged before the request is sent. There's room for failure between this line and the GET request being sent.

laurafitzgerald:
@davidffrench @camilamacedo86 I've rebased this pr and pushed the new changes.
I have verified locally that `make create-app` and `make delete-app` are working correctly.
It is also possible to update the Application Name value in the App CR and this will update the name in the service. You can verify all of these are working via the Mobile Security Service UI. I will update the verification steps in the description now to reflect that.
I believe that since starting the test work has been merged. The CI build is failing on the tests but the container image listed in the description is up to date and available for use in verification.

davidffrench:
After talking to @laurafitzgerald , the issue I have seen is not related to this PR. Last change is fixing the tests and then this can be merged.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: It will be used to create the SDKConfigMap info. We need to fix it now. 
Without the route is not possible to add the ServiceURL into it.

pkg/controller/mobilesecurityservice/status.go
camilamacedo86: It will not work as should be. Note that you are subscribing the APP_SERVICE name and status with the proxy ones in the Status fields of the CR.

pkg/controller/mobilesecurityservice/status.go
camilamacedo86: @laurafitzgerald we need to keep it to try to avoid issues which make the reconcile be called again. Note that from one point to another the CR can be already changed then it will fail.

pkg/controller/mobilesecurityservice/status.go
camilamacedo86: I comment it above. 

> Regards item 4.
> 
> It is not a check. The impl will update the field ServiceStatus in the CR with the Status of the object returned by the API. In this way, is required to create a new Status for the Proxy Service and then do an impl to update it. Please, let me know if it makes sense for you and if not let's chat by BJ.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: I did comment over it above.

> The GetApiURL is used in the SDKConfigMap. Did you check the values on it? I think we need to keep an impl for it still return the data from the route as it was done before.

PR #83, opened 2019-05-16
Remove role and role binding since has no need for it.

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8968

## What
- Remove the role and role_binding and the commands which were using it
- Update readme

## Why
It is not a requirement has no need for we to spend time on it. 

## Verification Steps
 
Use this PR to install the operator and ensure that after run "make create-all" all will be working fine.

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


davidffrench:
@camilamacedo86 Could you explain why the role is no longer needed? Was it ever being used? If it was, what was it used for?

camilamacedo86:
I added it to allow install the MSS and its database in another namespace managed by another user. For example, a developer which is not a requirement. In order to make simple and we do not need to ensure that I just remove it.

PR #86, opened 2019-05-16
feat(MSS) : Allow the MSS CR be able to specify how much resources …

camilamacedo86:
## Motivation
A debate in the refined backlog meeting today

## What
Allow the MSS object define/specify how much of the resources required for it works well should have the DB used by it. 

## Why
- What is required for the object MSS make sense be an attribute for it
- Attend the requirement which is that all specs required for the MSS can be performed in its CR

## How
- Add the optional Database resources requirements for the MSS CR
- If the MSS CR decided not to define it them the default config values for the DB will be used.

## Verification Steps
- With this PR and the image `docker.io/cmacedo/mobile-security-service-operator:DB-SPECS-FOR-MSS`
- Check that all was installed with success after run `make create-all`
- Check the changes in the README.

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

 
Screenshot 2019-05-16 at 14 13 51

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -157,6 +157,30 @@ NOTE:
 * All values used in the default configuration are sourced from the config-map which is managed and created by the Operator. This config map will be created in the Operator namespace and its name is defined by the attribute `configMapName` in the link:./deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml[MobileSecurityService CR].
 * If the name of this ConfigMap is not specified then the name of the Mobile Security Service instance will be used instead.
 
+=======
+=== Database resources required by the Mobile Security Service
+
+In the link:./deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml[MobileSecurityService CR] you are able to define the quantity of resources required and its limits for the database which will be used by this service. Following the fields.
+
+[source,yaml]
+----
+  databaseMemoryLimit: "512Mi"
+  databaseMemoryRequest: "512Mi"
+  databaseStorageRequest: "1Gi"
+----
+
+NOTE: This values are optional and if be not defined the Database will be installed with your own default configurations specified in its link:./deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservicedb_cr.yaml[MobileSecurityServiceDB CR].
+
+=== Configure roles for non-privileged users
+
+By executing the following commands you will create the required roles which allow the <user> to create the Mobile Security Service Application and Database in their namespaces so it is not required to be logged in to the system:admin user. However, the Mobile Security Service Operator is cluster scoped and will still only accessible to the system:admin user.
+
+[source,shell]
+----
+$ oc create rolebinding developer-mobile-security-service-operator --role=mobile-security-service-operator --user=<user>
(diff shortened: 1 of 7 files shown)


camilamacedo86:
Hi @davidffrench,

Impl:
* The config regards the DB requirements/resources expected by the MSS Service are optional
* If the MSS CR has not this specified then DB will be installed with it owns specs. 
* The DB and MSS still able to working independently.

Explanation:

The DB has either these parameters. I think it is easier to understand it by reading the changes in the Readme. It is because they would like to define in the project all specs that are required for it works. (Add all config that is required for the Service works in the Service CR). 

Since we can say that the MSS Object Class can have as an attribute of this class the requirements required for it works well and what it is expecting have regarded the DB then it shows fine for me.

davidffrench:
Apologies @camilamacedo86 , I still don't understand this change. If the DB CR didn't exist when the MSS CR was instantiated and a DB CR was created automatically, then this would make sense. If this is not the case, I don't understand the need for these changes.

camilamacedo86:
Hi @davidffrench,
Following a better explanation. Let me know if it will make sense if not we can chat by BJ.

## WHAT: 
Allow the users/RHMI/MDC update all possible config requires for the Service in just its CR

## HOW:
When the DB which will be used by the service be installed/config it will use the definitions made as requirements in the Service. If the specs were not defined in the Service then the default values described in the DB CR will be used. 

**NOTE:** All the other values which are requirements for the service and should be used in the DB already works in this way. See 

## WHY
Attend the requirements made by the architecture which is: 
* All config required for the Service works as for example the size of the DB which it will be used be it should be defined in the Service. 
* All required/possible config which could be used in the Service be in the Service
* Do not have the need to update/manage the config/attributes values of 2 CRs. They would like to have the need to update just one. 

## SOLUTION

IMHO, it is conceptually 100% fine since we have a Class Object called Service and we are defining here as attributes of this class all requirements that it can have, for example:

> databaseStorageRequest: What is the size that Service requires in the DB which will be used by it?

davidffrench:
@camilamacedo86 I understand what you are saying. However, I still do not understand the reasoning behind it.

If when creating the service CR, the DB cr was created by the operator automatically, I would understand the reason behind this. Since this is not the case and they are two independent CRs, I believe this is adding unnecessary complexity. It makes sense to have the db name, user and pwd as it needs to know what database to connect too. Other than these values, the service CR should not know anything else about the database or the database config.

camilamacedo86:
Hi @davidffrench,

If you read the comments https://github.com/aerogear/unifiedpush-operator/pull/6#pullrequestreview-236120406 you will see that the only reason for they create the DB object as part of the Service Object is that they would like to have the option to change this values in the Service CR.

davidffrench:
@camilamacedo86 We should not be mixing our two CRs together though. From https://github.com/aerogear/unifiedpush-operator/pull/6#pullrequestreview-236120406, since both the service and DB are created form the same CR, it makes sense to have the database values exposed there. I do not see a valid reason for us to be adding the resource values for one CR to a completely different CR.

camilamacedo86:
Oks, I am closing this one .

PR #88, opened 2019-05-16
Doing fixes and refacs in the code

camilamacedo86:
## Motivation
- A comment raised by a colleague across the teams. 
- https://github.com/aerogear/mobile-security-service-operator/pull/81
- https://issues.jboss.org/browse/AEROGEAR-9231
- https://issues.jboss.org/browse/AEROGEAR-9213
- https://issues.jboss.org/browse/AEROGEAR-9156

## What

- Fix impl to update the status of the Services in the CR. (introduced at https://github.com/aerogear/mobile-security-service-operator/pull/81)
- Fix impl to create the values of the SDKConfigMap  (introduced at https://github.com/aerogear/mobile-security-service-operator/pull/81)
- Keep the reconcile calls centralized in the reconcile func in order to make the code more understandable, clean and follow the same standard proposed in the operator-framework docs and examples.
- NOT do the re-queue when an object is created/updated always as it was since the reconcile calls were not centralized as described in the item above.
- Remove return of error since has no error scenario in this case that could be returned
- Fix code format by using `make fmt`
-  add setup to githooks
- fix export of variables locally for the run-local command.

## Why
- Solve issues
- Good Practices
- Make the code more understandable.

## Verification Steps
* Use the image `docker.io/cmacedo/mobile-security-service-operator:REF-RECONCILE`
* Run `make create-all` and ensure that all will be installed and get running with success.

### Check the  MSS CR Status
* Go to `Resources >> Other Resources` and check that has a proxyServiceName and  proxyServiceStatus for the MSS CR. Check if the names of the ServiceName and of ProxyServiceName are correct.
Screenshot 2019-05-21 at 09 18 38

### Check the API URL in the SDKConfigMap

* Create the namespace `mobile-security-service-apps` 
* Create the app by using `make create-app`
* Check if the SDKConfigMap of this app has as URL the route + API_ENDPOINT
* Check that the app was created in the Service

Screenshot 2019-05-21 at 09 19 56
Screenshot 2019-05-21 at 09 19 29
Screenshot 2019-05-21 at 09 19 18
Screenshot 2019-05-21 at 09 19 56

### Check delete an app
* Run `make delete-all`
* Check that the app was removed from the Service

Screenshot 2019-05-21 at 09 23 04

**NOTE:** The delete all worked perfectly. 

Screenshot 2019-05-21 at 09 23 52

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

### DIFF
--- a/pkg/controller/mobilesecurityservice/controller.go
+++ b/pkg/controller/mobilesecurityservice/controller.go
@@ -88,47 +88,42 @@ type ReconcileMobileSecurityService struct {
 }
 
 //Update the factory object and requeue
-func (r *ReconcileMobileSecurityService) update(obj runtime.Object, reqLogger logr.Logger) (reconcile.Result, error) {
+func (r *ReconcileMobileSecurityService) update(obj runtime.Object, reqLogger logr.Logger) error {
 	err := r.client.Update(context.TODO(), obj)
 	if err != nil {
 		reqLogger.Error(err, "Failed to update Spec")
-		return reconcile.Result{}, err
+		return err
 	}
 	reqLogger.Info("Spec updated - return and create")
-	return reconcile.Result{Requeue: true}, nil
+	return nil
 }
 
 //Create the factory object and requeue
-func (r *ReconcileMobileSecurityService) create(instance *mobilesecurityservicev1alpha1.MobileSecurityService, reqLogger logr.Logger, kind string) (reconcile.Result, error) {
-	obj, err := r.buildFactory(reqLogger, instance, kind)
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-
+func (r *ReconcileMobileSecurityService) create(instance *mobilesecurityservicev1alpha1.MobileSecurityService, reqLogger logr.Logger, kind string) error {
+	obj := r.buildFactory(reqLogger, instance, kind)
 	reqLogger.Info("Creating a new ", "kind", kind, "Namespace", instance.Namespace)
-	err = r.client.Create(context.TODO(), obj)
+	err := r.client.Create(context.TODO(), obj)
 	if err != nil {
 		reqLogger.Error(err, "Failed to create new ", "kind", kind, "Namespace", instance.Namespace)
-		return reconcile.Result{}, err
 	}
 	reqLogger.Info("Created successfully - return and create", "kind", kind, "Namespace", instance.Namespace)
-	return reconcile.Result{Requeue: true}, nil
+	return err
 }
 
 // buildFactory will return the resource according to the resource defined
-func (r *ReconcileMobileSecurityService) buildFactory(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService, resource string) (runtime.Object, error) {
+func (r *ReconcileMobileSecurityService) buildFactory(reqLogger logr.Logger, instance *mobilesecurityservicev1alpha1.MobileSecurityService, resource string) runtime.Object {
 	reqLogger.Info("Check "+resource, "into the namespace", instance.Namespace)
 	switch resource {
 	case CONFIGMAP:
-		return r.buildConfigMap(instance), nil
+		return r.buildConfigMap(instance)
 	case DEEPLOYMENT:
-		return r.buildDeployment(instance), nil
+		return r.buildDeployment(instance)
 	case PROXY_SERVICE:
-		return r.buildProxyService(instance), nil
+		return r.buildProxyService(instance)
 	case APPLICATION_SERVICE:
-		return r.buildApplicationService(instance), nil
+		return r.buildApplicationService(instance)
 	case ROUTE:
-		return r.buildRoute(instance), nil
+		return r.buildRoute(instance)
 	default:
 		msg := "Failed to recognize type of object" + resource + " into the Namespace " + instance.Namespace
 		panic(msg)
@@ -168,40 +163,55 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 
 	//Check specs
 	if !hasMandatorySpecs(instance, reqLogger) {
-		return reconcile.Result{Requeue: true}, nil
+		//Stop to reconcile since has not the mandatory specs
+		return reconcile.Result{}, nil
 	}
 
 	//Check if ConfigMap for the app exist, if not create one.
 	if _, err := r.fetchConfigMap(reqLogger, instance); err != nil {
-		return r.create(instance, reqLogger, CONFIGMAP)
+		if err := r.create(instance, reqLogger, CONFIGMAP); err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
 	//Check if Deployment for the app exist, if not create one
 	deployment, err := r.fetchDeployment(reqLogger, instance)
 	if err != nil {
-		return r.create(instance, reqLogger, DEEPLOYMENT)
+		if err := r.create(instance, reqLogger, DEEPLOYMENT); err != nil {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{Requeue: true}, nil
 	}
 
 	reqLogger.Info("Ensuring the Mobile Security Service deployment size is the same as the spec")
 	size := instance.Spec.Size
 	if *deployment.Spec.Replicas != size {
+		//Set size of the deployment spec
 		deployment.Spec.Replicas = &size
-		return r.update(deployment, reqLogger)
-	}
-
-	//Check if Service for the app exist, if not create one
-	if _, err := r.fetchService(reqLogger, instance, utils.PROXY_SERVICE_INSTANCE_NAME); err != nil {
-		return r.create(instance, reqLogger, PROXY_SERVICE)
+		if err := r.update(deployment, reqLogger); err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
 	//Check if Service for the app exist, if not create one
 	if _, err := r.fetchService(reqLogger, instance, utils.APPLICATION_SERVICE_INSTANCE_NAME); err != nil {
-		return r.create(instance, reqLogger, APPLICATION_SERVICE)
+		if err := r.create(instance, reqLogger, APPLICATION_SERVICE); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	//Check if Proxy Service for the app exist, if not create one
+	if _, err := r.fetchService(reqLogger, instance, utils.PROXY_SERVICE_INSTANCE_NAME); err != nil {
+		if err := r.create(instance, reqLogger, PROXY_SERVICE); err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
 	//Check if Route for the Service exist, if not create one
 	if _, err := r.fetchRoute(reqLogger, instance); err != nil {
-		return r.create(instance, reqLogger, ROUTE)
+		if err := r.create(instance, reqLogger, ROUTE); err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
 	//Update status for ConfigMap
@@ -217,12 +227,12 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 	}
 
 	//Update status for Proxy Service
-	proxyServiceStatus, err := r.updateServiceStatus(reqLogger, request, utils.PROXY_SERVICE_INSTANCE_NAME)
+	proxyServiceStatus, err := r.updateProxyServiceStatus(reqLogger, request)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
-	applicationServiceStatus, err := r.updateServiceStatus(reqLogger, request, utils.APPLICATION_SERVICE_INSTANCE_NAME)
+	applicationServiceStatus, err := r.updateAppServiceStatus(reqLogger, request)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
@@ -235,7 +245,6 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 
 	//Update status for App
 	if err := r.updateStatus(reqLogger, configMapStatus, deploymentStatus, proxyServiceStatus, applicationServiceStatus, routeStatus, request); err != nil {
-
 		return reconcile.Result{}, err
 	}
 
--- a/pkg/controller/mobilesecurityservice/fakeclient_test.go
+++ b/pkg/controller/mobilesecurityservice/fakeclient_test.go
@@ -1,0 +1,29 @@
+package mobilesecurityservice
+
+import (
+	mobilesecurityservicev1alpha1 "github.com/aerogear/mobile-security-service-operator/pkg/apis/mobilesecurityservice/v1alpha1"
+	routev1 "github.com/openshift/api/route/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"testing"
+)
+
+//buildReconcileWithFakeClientWithMocks return reconcile with fake client, schemes and mock objects
+func buildReconcileWithFakeClientWithMocks(objs []runtime.Object, t *testing.T) *ReconcileMobileSecurityService {
+	s := scheme.Scheme
+
+	// Add route Openshift scheme
+	if err := routev1.AddToScheme(s); err != nil {
+		t.Fatalf("Unable to add route scheme: (%v)", err)
+	}
+
+	s.AddKnownTypes(mobilesecurityservicev1alpha1.SchemeGroupVersion, &mobilesecurityservicev1alpha1.MobileSecurityService{})
+
+	// create a fake client to mock API calls with the mock objects
+	cl := fake.NewFakeClient(objs...)
+
+	// create a ReconcileMobileSecurityService object with the scheme and fake client
+	return &ReconcileMobileSecurityService{client: cl, scheme: s}
+}
--- a/pkg/controller/mobilesecurityservice/mocks_test.go
+++ b/pkg/controller/mobilesecurityservice/mocks_test.go
@@ -1,0 +1,52 @@
+package mobilesecurityservice
+
+import (
+	mobilesecurityservicev1alpha1 "github.com/aerogear/mobile-security-service-operator/pkg/apis/mobilesecurityservice/v1alpha1"
+	"github.com/aerogear/mobile-security-service-operator/pkg/utils"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+)
+
+var (
+	mssInstance = mobilesecurityservicev1alpha1.MobileSecurityService{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "mobile-security-service-app",
+			Namespace: "mobile-security-service-operator",
+		},
+		Spec: mobilesecurityservicev1alpha1.MobileSecurityServiceSpec{
+			Size:                    1,
+			MemoryLimit:             "512Mi",
+			MemoryRequest:           "512Mi",
+			ClusterProtocol:         "http",
+			ConfigMapName:           "mss-config",
+			RouteName:               "route",
+			SkipNamespaceValidation: true,
+		},
+	}
+
+	mssInstance2 = mobilesecurityservicev1alpha1.MobileSecurityService{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "mobile-security-service-app",
+			Namespace: "mobile-security-service-operator-2",
+		},
+		Spec: mobilesecurityservicev1alpha1.MobileSecurityServiceSpec{
+			Size:                    1,
+			MemoryLimit:             "512Mi",
+			MemoryRequest:           "512Mi",
+			ClusterProtocol:         "http",
+			ConfigMapName:           "mss-config",
+			RouteName:               "route",
+			SkipNamespaceValidation: true,
+		},
+	}
+
+	configMap = corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      utils.GetConfigMapName(&mssInstance),
+			Namespace: mssInstance.Namespace,
+			Labels:    getAppLabels(mssInstance.Name),
+		},
+		Data: getAppEnvVarsMap(&mssInstance),
+	}
+)
--- a/pkg/controller/mobilesecurityservice/status.go
+++ b/pkg/controller/mobilesecurityservice/status.go
@@ -32,7 +32,6 @@ func (r *ReconcileMobileSecurityService) updateStatus(reqLogger logr.Logger, con
 
 	// Update CR with the AppStatus == OK
 	if !reflect.DeepEqual(status, instance.Status.AppStatus) {
-
 		// Set the data
 		instance.Status.AppStatus = status
 
@@ -64,6 +63,12 @@ func (r *ReconcileMobileSecurityService) updateConfigMapStatus(reqLogger logr.Lo
 
 	// Update ConfigMap Name
 	if configMapStatus.Name != instance.Status.ConfigMapName {
+		// Get the latest version of the CR in order to try to avoid errors when try to update the CR
+		instance, err := r.fetchInstance(reqLogger, request)
+		if err != nil {
+			return nil, err
+		}
+
 		// Set the data
 		instance.Status.ConfigMapName = configMapStatus.Name
 
@@ -95,7 +100,7 @@ func (r *ReconcileMobileSecurityService) updateDeploymentStatus(reqLogger logr.L
 
 	// Update the Deployment Name and Status
 	if deploymentStatus.Name != instance.Status.DeploymentName || !reflect.DeepEqual(deploymentStatus.Status, instance.Status.DeploymentStatus) {
-		// Get the latest version of the CR
+		// Get the latest version of the CR in order to try to avoid errors when try to update the CR
 		instance, err := r.fetchInstance(reqLogger, request)
 		if err != nil {
 			return nil, err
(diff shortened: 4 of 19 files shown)


camilamacedo86:
Hi @Folks,

All examples show doing the re-queue after create/update. I'd like to understand a few things better before we move forward here with its team members.

grdryn:
@camilamacedo86 is there somewhere you could point me to where this is suggested or mentioned as a  good practice? It's something we've been thinking about recently, but I'm not sure which way it should be.

craicoverflow:
Hi @camilamacedo86 I noticed you are not fully following the conventional commit message standard.

`ref(reconcile create/update) : ` - ref means refactor yes? In a previous commit message you did:

`refactory(xx..` and another you uses `refac(xx...`

Could you please choose one? Preferably `refactor`.

Screenshot from 2019-05-17 09-52-03

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -192,20 +195,21 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
+		if err := r.create(instance, CONFIGMAP, serviceAPI, reqLogger, request) ; err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
 	//Check if App is Bind in the REST Service, if not then bind it
 	if app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger); err == nil {
 		//Update the name by the REST API
-		if app.AppName != instance.Spec.AppName {
+		if app.ID != "" && app.AppName != instance.Spec.AppName {
```
craicoverflow: Why was this additional check added?

pkg/controller/mobilesecurityservicedb/controller.go
```diff
@@ -83,31 +83,31 @@ type ReconcileMobileSecurityServiceDB struct {
...
 	reqLogger.Info("Creating a new ", "kind", kind, "Namespace", instance.Namespace)
 	err = r.client.Create(context.TODO(), obj)
 	if err != nil {
 		reqLogger.Error(err, "Failed to create new ", "kind", kind, "Namespace", instance.Namespace)
-		return reconcile.Result{}, err
+		return err
 	}
 	reqLogger.Info("Created successfully - return and create", "kind", kind, "Namespace", instance.Namespace)
-	return reconcile.Result{Requeue: true}, nil
+	return nil
```
craicoverflow: You could just return `err` here and remove the return from line 107.

pkg/controller/mobilesecurityservicedb/controller.go
```diff
@@ -174,25 +174,37 @@ func (r *ReconcileMobileSecurityServiceDB) Reconcile(request reconcile.Request)
 		reqLogger.Info("Checking for service instance ...")
 		serviceInstance := &mobilesecurityservicev1alpha1.MobileSecurityService{}
 		r.client.Get(context.TODO(), types.NamespacedName{Name: utils.SERVICE_INSTANCE_NAME, Namespace: operatorNamespace}, serviceInstance)
-		return r.create(instance, serviceInstance, DEEPLOYMENT, reqLogger)
+
+		// Create the deployment
+		if err := r.create(instance, serviceInstance, DEEPLOYMENT, reqLogger) ; err != nil {
```
craicoverflow: I see a space between `)` and `;`.

craicoverflow: I see this in several places. You should run `make fmt` to format the code.

grdryn: I guess a check for that could be added to the CI job, right? If it ran the formatting command and there were changes it chould fail the job.

craicoverflow: Yep there should be. We have one in the `mobile-security-service` project. We also have it as part of the pre-commit hook there to speed up catching this.

davidffrench: +1 @grdryn and I believe we have this in our Go server already

pkg/controller/mobilesecurityserviceapp/controller.go
laurafitzgerald: I also made the same change in my pr. the check is if an app is returned (id != "") and the appname is not the same as specified by the cr (i.e. the cr is update) then go ahead and update the app in the service. I believe the change is correct here.

pkg/controller/mobilesecurityservicedb/controller.go
laurafitzgerald: Looks like there is a JIRA already for that: https://issues.jboss.org/browse/AEROGEAR-9213

pkg/controller/mobilesecurityservicedb/controller.go
camilamacedo86: I could not. It will not build. 

Screenshot 2019-05-19 at 09 30 46

pkg/controller/mobilesecurityservicedb/controller.go
camilamacedo86: Yes, we have a JIRA for it and I think we should use this one to fix the format in the project.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: Hi @craicoverflow the check was added to ensure that the PATH request will just be called if the app is created in the REST Service

camilamacedo86:
HI @laurafitzgerald and @craicoverflow,

It is rebased with the master and a new image was build/pushed. Please, feel free to check.
NOTE: To rebase with the master a need fixed a bug introduced in the update status of the services in the CR. I got that is not clear for all what/where/why status we are updating here. I will schedule a BJ to share this knowledge.

camilamacedo86:
Hi @grdryn, 

> @camilamacedo86 is there somewhere you could point me to where this is suggested or mentioned as a good practice? It's something we've been thinking about recently, but I'm not sure which way it should be.

This PR changed in order to address what is required at this moment. However, the change here related to good practice was related to not "requeue" when it is not necessary. What document are you looking for as proof that is not a good practice perform many actions when they are not required at all? Or that is a good practice make the code more understandble for example?  
-

c/c @aliok

pkg/controller/mobilesecurityserviceapp/controller.go
craicoverflow: Thanks for clarifying.

pkg/controller/mobilesecurityservicedb/controller.go
craicoverflow: You would have to replace `return nil` with `return err`, and remove `return err` from inside the `if` statement.

pkg/utils/utils.go
```diff
@@ -16,9 +17,20 @@ import (
 const APP_NAMESPACE_ENV_VAR = "APP_NAMESPACES"
 const PROXY_SERVICE_INSTANCE_NAME = "mobile-security-service-proxy"
 const APPLICATION_SERVICE_INSTANCE_NAME = "mobile-security-service-application"
+const ENDPOINT_API = "/api"
 
 var log = logf.Log.WithName("mobile-security-service-operator.utils")
 
+//Return the router host URL
+func getServiceURL(route *routev1.Route, serviceInstance *mobilesecurityservicev1alpha1.MobileSecurityService) string{
+	return serviceInstance.Spec.ClusterProtocol + "://" + route.Status.Ingress[0].Host
```
craicoverflow:
```suggestion
return fmt.Sprintf("%v://%v", serviceInstance.Spec.ClusterProtocol, route.Status.Ingress[0].Host)
```

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -168,40 +164,56 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
...
 	}
 
 	//Check if Deployment for the app exist, if not create one
 	deployment, err := r.fetchDeployment(reqLogger, instance)
 	if err != nil {
-		return r.create(instance, reqLogger, DEEPLOYMENT)
+		if err := r.create(instance, reqLogger, DEEPLOYMENT) ; err != nil {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{Requeue: true}, nil
```
craicoverflow: https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#reconcile-loop

The documentation says:

```
// Requeue for any reason other than error
return reconcile.Result{Requeue: true}, nil
```

But you are re-queuing after an error here.

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -168,40 +164,56 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 
 	//Check specs
 	if !hasMandatorySpecs(instance, reqLogger) {
-		return reconcile.Result{Requeue: true}, nil
+		//Stop to reconcile since has not the mandatory specs
+		return reconcile.Result{}, nil
```
craicoverflow: https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#reconcile-loop

```go
// Reconcile failed due to error - requeue
```

What constitutes an error? Here you are not re-queuing but the docs say you should requeue when there is not an error.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -171,48 +170,56 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
 
-	//Check if the app exists, if yes update
-	if app.AppName != "" && app.AppName != instance.Spec.AppName {
-		app.AppName = instance.Spec.AppName
-		//Check if App was update with success
-		//TODO if the app is found but it is soft deleted, we need to remove the soft delete
-		err := service.UpdateAppNameByRestAPI(serviceAPI, app, reqLogger)
-		if err != nil {
+	//Check if App is Bind in the REST Service, if not then bind it
+	if app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger); err == nil {
```
laurafitzgerald: @camilamacedo86 some of the changes I made here are linked to changes that I made inside the client service. I see that you have reverted them. Can you tell me why you've changed them back?

grdryn:
Hi @camilamacedo86, don't worry, I was not saying that this is wrong, or that that you shouldn't do it. It's just that this was something I was wondering about before you even created the PR, and your description previously said that it was a good practice not to requeue after creating/updating resources. Given that the reference example _does_ requeue after creating resources, I was just asking where it was documented that it's better not to do it, so that I can read that and understand better. :smiley: 

> Hi @grdryn,
> 
> > @camilamacedo86 is there somewhere you could point me to where this is suggested or mentioned as a good practice? It's something we've been thinking about recently, but I'm not sure which way it should be.
> 
> I am not sure if I understood your point here. My sincere apologies if I could not make the propose clear at the first moment. In order to try to clarify, following the proposal of the main changes performed previously.
> 
>     1. Keep the reconcile calls centralized in the reconcile func in order to make the code more understandable, clean and follow the same `standard` proposed in the operator-framework docs and examples.
> 
>     2. NOT  do the re-queue when an object is created/updated `always` as it was since the reconcile calls were not centralized as described in the item above.

Point 2 here is the part that I'm not clear about. The memcached-operator example _does_ requeue after successfully creating objects (and it's in the Reconcile method): https://github.com/operator-framework/operator-sdk/blob/master/example/memcached-operator/memcached_controller.go.tmpl#L122..L123

grdryn:
> @grdryn
> 
> Is [this](https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#reconcile-loop) what you are looking for?

That doesn't say one way or another whether a requeue should be done after creating or updating a resource. It does, however, link to [this](https://github.com/operator-framework/operator-sdk/blob/master/doc/user/client.md#example-usage) page, and in the example there a requeue is done every time a create or update is done. My understanding of the premise of this PR is that we're now saying that's less good?

pkg/controller/mobilesecurityservice/fakeclient_test.go
```diff
@@ -0,0 +1,31 @@
...
+//buildReconcileWithFakeClientWithMocks return reconcile with fake client, schemes and mock objects
+func buildReconcileWithFakeClientWithMocks(objs []runtime.Object, t *testing.T) *ReconcileMobileSecurityService {
+	s := scheme.Scheme
+
+	// Add route Openshift scheme
+	if err := routev1.AddToScheme(s); err != nil {
+		t.Fatalf("Unable to add route scheme: (%v)", err)
+	}
+
+	s.AddKnownTypes(mobilesecurityservicev1alpha1.SchemeGroupVersion, &mobilesecurityservicev1alpha1.MobileSecurityService{})
```
craicoverflow: in my opinion we should have this function in each `_test.go` file. Each controller requires different known types to be added. so in the database controller test, we would need to replace this with:

```go
s.AddKnownTypes(mobilesecurityservicev1alpha1.SchemeGroupVersion, &mobilesecurityservicev1alpha1.MobileSecurityServiceDB{})
```

craicoverflow: It also requires a different return type.

camilamacedo86:
Hi @grdryn,

It was my first understanding of this [example](https://github.com/operator-framework/operator-sdk/blob/master/example/memcached-operator/memcached_controller.go.tmpl#L122..L123)., which was that we should do the re-queue always. However, if you check it has just one object and the requeue is just made because of the second check which will verify the qt of Replicas/Size that the deployment has. 

So, if you check this PR we are doing the same in this case where it is required. However, why is required re-start the loop(reconcile) again when the configMap is created for example? Has no need in the impl for that, do not you agree? What check we will do with it which requires the reconcile do all again? 

Note that the Memcached [example](https://github.com/operator-framework/operator-sdk/blob/master/example/memcached-operator/memcached_controller.go.tmpl#L122..L123)  also could be changed by after creating the Deployment to a get to get the Deployment created. However, since it has just one object and I assume that they would like to show how to re-queue it is made at this way. 

Is it make sense? If not, please let schedule a BJ to chat over, It will be easy.

pkg/controller/mobilesecurityservice/fakeclient_test.go
camilamacedo86: HI @craicoverflow, 

But it is inside of the dir pkg/controller/mobilesecurityservice/. You will not be able to use it for the DB. You should do another file like this one with the impl required for the DB. Do not you agree? We could after to all check if has a better way to centralize the impl for all, but so far I think is good we keep one for each.

grdryn:
@camilamacedo86 Thanks for the explanation of the motivation/background behind this PR! :smiley:

> It was my first understanding of this example., which was that we should do the re-queue always. However, if you check it has just one object and the requeue is just made because of the second check which will verify the qt of Replicas/Size that the deployment has.

Good point! However, here's another example in that same function of another requeue, but where the updated resource (`found`) isn't used again after it's updated. I think this is probably done to be consistent, and to emphasise the pattern: if you create or update something, requeue.

It seems like either way is fine, and it's just down to personal preference. My preference (and what we'll likely do in the unifiedpush-operator) is to always requeue, even if we think that we won't need it. That way, it will be easier for someone to add additional logic in future without having to worry if a requeue was done before their code, and/or if they need to add one after they've updated something.

pkg/controller/mobilesecurityservice/fakeclient_test.go
craicoverflow: Yes sorry I did not notice that it was within this package. I don't believe there is any reason to create a new file for this then.

camilamacedo86:
Hi @grdryn, 

Over your last comment which follows. 

> It seems like either way is fine, and it's just down to personal preference. My preference (and what we'll likely do in the unifiedpush-operator) is to always requeue, even if we think that we won't need it. That way, it will be easier for someone to add additional logic in future without having to worry if a requeue was done before their code, and/or if they need to add one after they've updated something.

**Is a good practice to make a loop be performed with many checks unnecessary which will spend more resources without any need for that? I don't think so. In this way, I do see the logic/reason to do the requeue always**.

pkg/controller/mobilesecurityservice/controller.go
camilamacedo86: It not after, it is when is no faced an error.
Just the deployment is required to be re-queue because of the size/replicas check.

pkg/controller/mobilesecurityservice/controller.go
camilamacedo86: IMPORTANT: It is not an error because the object was not created/updated. 

If the Specs are wrong means that to fix is required delete and re-apply the CR. So, in this case, for our implementation here no make sense continue doing the loop until this check 4ever and the reconcile should stop.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: I just remove the call to create/bind app which was added and it is redundant. The app will be created inside of this block, so why call the impl again?

pkg/controller/mobilesecurityservice/fakeclient_test.go
camilamacedo86: The idea is just to make implicitly that the code is used by the status and controller.

pkg/controller/mobilesecurityservice/controller.go
craicoverflow: Agreed.

pkg/controller/mobilesecurityservice/controller.go
craicoverflow: When I say "after an error" I mean when you get an error.

```go
if err != nil {
	if err := r.create(instance, reqLogger, DEEPLOYMENT) ; err != nil {
	}
    // an error has occurred, but you are re-queueing which contradicts the docs.
    return reconcile.Result{Requeue: true}, nil
}
```

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -171,48 +170,56 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
+		return reconcile.Result{}, err
 	}
 
-	app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger)
+	// Get the Public Service API URL which will be used to build the SDKConfigMap json
+	publicServiceURLAPI := utils.GetPublicServiceAPIURL(route, mssInstance)
 
-	if err != nil {
-		return reconcile.Result{Requeue: true}, err
+	//Check if ConfigMap for the app exist, if not create one.
```
laurafitzgerald: What are we using the route in the config map for?

craicoverflow: The route is needed to make an API request to delete the app from the database.

laurafitzgerald: We don't use the route anymore. We use an internal service so that the operator can bypass the oauth proxy container.

craicoverflow:
Dismissed approval as there is still outstanding queries.

pkg/controller/mobilesecurityserviceapp/controller.go
craicoverflow: Sorry I looked more closely. The route appears to be used in the config map as this information is used by the app, which requires the public API URL in order to make a request on startup to `/init`.

```sh
~/go/src/github.com/aerogear/mobile-security-service-operator AEROGEAR-9214
$ oc describe cm example-appname-sdk
Name:         example-appname-sdk
Namespace:    mobile-security-service-apps
Labels:       app=mobilesecurityservice
              mobilesecurityserviceapp_cr=mobile-security-service-app
Annotations:  <none>

Data
====
SDKConfig:
----
{
  "version": "1.0.0",
  "name": "example-appname",
  "namespace": "mobile-security-service-apps",
  "clientId": "appid.example.com",
  "services": [
    {
      "id": "security",
      "name": "security",
      "type": "security",
      "url": "http://route-mobile-security-service.192.168.99.100.nip.io/api",
      "config": {
        "mobile-security-server-url": "http://route-mobile-security-service.192.168.99.100.nip.io/api"
      }
    }
  ]
}

```

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -171,48 +170,56 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
+			return reconcile.Result{}, err
+		}
 	}
-	//if no error found, i.e. request completed successfully
 
-	//Check if the app exists, if yes update
-	if app.AppName != "" && app.AppName != instance.Spec.AppName {
-		app.AppName = instance.Spec.AppName
-		//Check if App was update with success
-		//TODO if the app is found but it is soft deleted, we need to remove the soft delete
```
laurafitzgerald: I believe this TODO is still valid. Can we leave it in?

pkg/controller/mobilesecurityservice/controller.go
camilamacedo86: HI @craicoverflow,

I think you did not get it. It is like all other impls. Following the code with the explanation.

```
//Check if Deployment for the app exist, if not create one
deployment, err := r.fetchDeployment(reqLogger, instance) // IT WILL SEARCH
	if err != nil {  // IF HAS ERROR MEANS THAT THE OBJECT IS NOT CREATED
		if err := r.create(instance, reqLogger, DEEPLOYMENT) ; err != nil { // THEN, CREATE THE OBJ
			return reconcile.Result{}, err  // IF WAS NOT POSSIBLE CREATE THEN RECONCILE WITH ERR
		}
		return reconcile.Result{Requeue: true}, nil // IF WAS CREATED, THEN RECONCILE WITH SUCCESS.
	}
```

pkg/controller/mobilesecurityservice/mocks_test.go
```diff
@@ -0,0 +1,66 @@
...
+			MemoryLimit:             "512Mi",
+			MemoryRequest:           "512Mi",
+			ClusterProtocol:         "http",
+			ConfigMapName:           "mss-config",
+			RouteName:               "route",
+			SkipNamespaceValidation: true,
+		},
+	}
+
+	route = routev1.Route{
```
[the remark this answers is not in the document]
camilamacedo86: Tks for the catch.

pkg/controller/mobilesecurityservice/status.go
```diff
@@ -141,14 +151,54 @@ func (r *ReconcileMobileSecurityService) updateServiceStatus(reqLogger logr.Logg
...
+	}
+	// Get the Service Object
+	serviceStatus, err := r.fetchService(reqLogger, instance, utils.PROXY_SERVICE_INSTANCE_NAME)
+	if err != nil {
+		reqLogger.Error(err, "Failed to get Proxy Service for Status", "MobileSecurityService.Namespace", instance.Namespace, "MobileSecurityService.Name", instance.Name)
+		return serviceStatus, err
+	}
+
+	// Update the Service Status and Name
+	if serviceStatus.Name != instance.Status.ProxyServiceName || !reflect.DeepEqual(serviceStatus.Status, instance.Status.ProxyServiceStatus) {
```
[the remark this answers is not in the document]
camilamacedo86: HI, @craicoverflow  we cannot do `serviceStatus.Status != instance.Status.ProxyServiceStatus` it is an object and because of it, we need to check the values of its fields with `reflect.DeepEqual`

pkg/controller/mobilesecurityservice/status.go
```diff
@@ -170,6 +220,11 @@ func (r *ReconcileMobileSecurityService) updateRouteStatus(reqLogger logr.Logger
 
 	// Update the Route Status and Name
 	if routeName != instance.Status.RouteName || !reflect.DeepEqual(route.Status, instance.Status.RouteStatus) {
```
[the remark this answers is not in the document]
camilamacedo86: HI, @craicoverflow,

It is an object because of it we are using `reflect.DeepEqual`.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: @laurafitzgerald,

Look the steps to verify as follows. 

> Check the API URL in the SDKConfigMap
> Create the namespace mobile-security-service-apps
> Create the app by using make create-app
> Check if the SDKConfigMap of this app has as URL the route + API_ENDPOINT

If you follow that you will understand.

pkg/controller/mobilesecurityservicedb/controller.go
camilamacedo86: Done

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: In reality no. But was good your comment because I saw that it is not working as it was after the incremental work. 
I will push the changes and I will do the comments. Please, let me know if is clear and if it is working for you.

camilamacedo86: Could you check now?

pkg/controller/mobilesecurityserviceapp/controller.go
laurafitzgerald: Thanks for the information. It's probably seperate to this pr but I believe we only need to provide this section 
```
{
      "id": "security",
      "name": "security",
      "type": "security",
      "url": "http://route-mobile-security-service.192.168.99.100.nip.io/api",
      "config": {
        "mobile-security-server-url": "http://route-mobile-security-service.192.168.99.100.nip.io/api"
      }
``` as the MDC will gather all the service specific information and build mobile-services.json. 

Again I don't know if this falls within the scope of this pr so feel free to ignore it for now.

pkg/controller/mobilesecurityserviceapp/controller.go
laurafitzgerald: @camilamacedo86 it still looks like it's removed.

laurafitzgerald: Sorry I see it here now https://github.com/aerogear/mobile-security-service-operator/pull/88/files#diff-f0c3b61d21a85c4c894842a29ca22af2R211

But I dont think the logic there will remove the soft delete value. Can we do a bluejeans about it tomorrow?

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: We can do the change here if you wish. It is easy. Really tks for the help in the review. 
Are you sure that we should change it? According to the requirements, it should be the json of the SDK.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: IF we call the POST in the SERVICE it will bind the app always. If it was soft deleted before then the deleted_at will be updated with null. We can do the BJ if you wish no problem at all.

pkg/utils/utils.go
camilamacedo86: 👍

camilamacedo86:
Hi @laurafitzgerald @craicoverflow,

Note that I added images to prove that all is working perfectly with this PR. Please, we need to approve and merge it.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: We can check it after. Please, do not forgot about it.

laurafitzgerald: @camilamacedo86 let's not do the change here but check with Mobile Services Team what they expect to recieve and create a task to update it.

pkg/controller/mobilesecurityserviceapp/controller.go
laurafitzgerald: Ok. Thanks for clarifying that. Happy to resolve this one.

pkg/controller/mobilesecurityserviceapp/controller.go
laurafitzgerald: Resolving this one as we discussed it in planning and there is a task to follow up on this.

laurafitzgerald:
**Check the MSS CR Status**
- [x] Check if the names of the ServiceName and of ProxyServiceName are correct.
**Check the API URL in the SDKConfigMap**
- [x] Check if the SDKConfigMap of this app has as URL the route + API_ENDPOINT
- [x] Check that the app was created in the Service
**Check delete an app**
- [x] Check that the app was removed from the Service

I have verified the functionality as shown above.
One note: There is an additional slash in the api endpoint

Screenshot 2019-05-21 at 12 25 19

PR #90, opened 2019-05-20
feat(add-metrics): add metrics to the operator :(INTLY-1455)

austincunningham:
## Motivation
https://issues.jboss.org/browse/INTLY-1455

## What
enable metrics for the mobile-security-service-operator

## Why
Required functionality for a managed service

## How
added a service monitor , and service and code to enable 

## Verification Steps
 
- start minishift as outline at https://github.com/aerogear/mobile-security-service-operator#minishift-installation-and-setup
- deploy the https://github.com/integr8ly/application-monitoring-operator 
  - Note some Grafana and Prometheus images won't pull so change the image registry to `registry.access.redhat.com` to install
- clone this repo and checkout this branch
- `cd mobile-security-service-operator`
- change the operator image to point at dev this line https://github.com/aerogear/mobile-security-service-operator/blob/4bad63f27c63cefd2241cb2ba6a6e0ce0129e437/deploy/operator.yaml#L19
to 
```
          image: quay.io/aerogear/mobile-security-service-operator:0.1.0-dev
```
- Install the operator
```bash
make create-all
```
- Check the Prometheus to see if metrics are being gathered 
Prometheus

## Checklist:

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -142,6 +142,22 @@ $ make delete-all
 
 == Configuration and Options
 
+=== Metrics
+
+The application-monitoring stack provisioned by the
+https://github.com/integr8ly/application-monitoring-operator[application-monitoring-operator]
+can be used to gather metrics from the operator here.  Once you have
+provisioned that (or the ServiceMonitor CRD at a minimum), it is configured by default with `make create-all` 
+you can run the following commands to configure it manually:
+
+[source,shell]
+----
+kubectl label namespace mobile-security-service monitoring-key=middleware
+kubectl create -f deploy/service_monitor.yaml
+kubectl create -f deploy/operator_service.yaml namespace mobile-security-service
+----
+
+
 === Oauth Configuration
 
 An Oauth Proxy container and the required configuration will be setup by default by the operator to provide authentication to the Mobile Security Service.
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aerogear/mobile-security-service-operator/pkg/apis"
 	"github.com/aerogear/mobile-security-service-operator/pkg/controller"
 	routev1 "github.com/openshift/api/route/v1"
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	"github.com/operator-framework/operator-sdk/pkg/metrics"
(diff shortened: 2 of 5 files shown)


README.adoc
```diff
@@ -142,6 +142,22 @@ $ make delete-all
 
 == Configuration and Options
 
+=== Metrics
+
+The application-monitoring stack provisioned by the
+https://github.com/integr8ly/application-monitoring-operator[application-monitoring-operator]
```
camilamacedo86: Should we use the integr8ly project? I think that we should use Prometheus only.

cmd/manager/main.go
```diff
@@ -72,13 +73,21 @@ func main() {
...
+		os.Exit(1)
+	}
 
 	//Create cmd Manager
 	//FIXME: We should not watch/cache all namespaces. However, the current version do not allow us pass the List of Namespaces.
 	// The impl to allow do it is done and merged in the master branch of the lib but not released in an stable version.
 	// See the PR which we are working on to update the deps and have this feature: https://github.com/operator-framework/operator-sdk/pull/1388
 	mgr, err := manager.New(cfg, manager.Options{
-		Namespace: "", 
+		Namespace: namespace,
```
camilamacedo86: It cannot be made. It will not allow the operator to work with other namespaces which are required to creating and delete the app.

README.adoc
austincunningham: No this is for the mobile-security-service as a managed service on integr8tly

cmd/manager/main.go
austincunningham: necessary as documented by the operator sdk https://github.com/operator-framework/operator-sdk/blob/master/doc/user/metrics/README.md

davidffrench: @austincunningham Camila is correct here, this will limit the Operator to only watch for resource events in one namespace. We need to watch for other namespaces due to the app CR which will be created in the MDC namespace.

camilamacedo86: Hi @austincunningham the code example in the docs is adding the namespace because it is a scoped operator. We cannot do that here and it will make the project stop to work and broken its functionalities, it will change with the future version of the operator-sdk where we will be able to tell to manager cache/watch a list of namespaces, so far it needs to have as value ""

camilamacedo86: See the following docs to have a better understanding. 
- https://github.com/operator-framework/getting-started#manager
- https://github.com/operator-framework/operator-sdk/blob/2e1c2bc3cdbe8c08a7b29e721e5ae635803adfbc/doc/operator-scope.md#L1

austincunningham: :+1:

craicoverflow:
Verified functionally that Prometheus is gathering information on the `mobile-security-service` project.

Screenshot from 2019-05-22 15-59-49

Screenshot from 2019-05-22 16-01-46

PR #97, opened 2019-05-21
fix(NamespaceCheck) : Fix namespace check for production and local env

camilamacedo86:
## Motivation
- https://issues.jboss.org/browse/AEROGEAR-9233
- https://issues.jboss.org/browse/AEROGEAR-8809

## What
- Remove skip from CR 
- Use instance namespace
- Check if it is running out of the cluster (unit test/local env) in the workarround valid namespace check

## Why
- Fix/improvements/Security

## How
- Check that it is a local env or a unit test and then return a default value for the Operator Namespace
- The isValidAppNamespace need to use the instance.Namespace
- Remove skip from CR

## Verification Steps
Test with the image: docker.io/cmacedo/mobile-security-service-operator:FIX-NAMESPACE-CHECK
 
1. Check that the operator was created with success in the cluster (make create-all)
2. Check the features to create and delete app ( Check ConfigMap and Service )
3. Check if the project will run locally to debug and etc with the command `make run-local` 
4. Check if it is passing in the CI

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -160,10 +160,10 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 	// We should not checked if the namespace is valid or not. It is an workaround since currently is not possible watch/cache a List of Namespaces
 	// The impl to allow do it is done and merged in the master branch of the lib but not released in an stable version. It should be removed when this feature be impl.
 	// See the PR which we are working on to update the deps and have this feature: https://github.com/operator-framework/operator-sdk/pull/1388
-	if isValidNamespace, err := utils.IsValidOperatorNamespace(request.Namespace, instance.Spec.SkipNamespaceValidation); err != nil || isValidNamespace == false {
+	if isValidNamespace, err := utils.IsValidOperatorNamespace(instance.Namespace); err != nil || isValidNamespace == false {
 		// Stop reconcile
 		operatorNamespace, _ := k8sutil.GetOperatorNamespace()
```
craicoverflow: I think you can add a test case to cover this condition now. Could you investigate this?

camilamacedo86: HI @craicoverflow I would like to move forward here to create a version for the demo. We can do it after since it is not a big priority at all. Wdyt?

PR #101, opened 2019-05-24
feat(status) : Add Status when the CR was applied in an invalid namespace (AEROGEAR-9267)

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9267

## What
- Add Invalid status

## Why
-Make clear that the namespace is invalid

## Verification Steps
Use the image:   docker.io/cmacedo/mobile-security-service-operator:AEROGEAR-9267
1. Install the oper
2. create an app in a namespace that is not valid and check the Status of this resource
3. try to create the MSS in an invalid namespace and check the status of this resource
4. try to create the DB in an invalid namespace and check the status of this resource

## Checklist:

## Progress

- [ ] Finished task
- [ ] TODO

### DIFF
--- a/pkg/controller/mobilesecurityservice/controller.go
+++ b/pkg/controller/mobilesecurityservice/controller.go
@@ -164,6 +164,12 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 		// Stop reconcile
 		operatorNamespace, _ := k8sutil.GetOperatorNamespace()
 		reqLogger.Error(err, "Unable to reconcile Mobile Security Service", "instance.Namespace", instance.Namespace, "isValidNamespace", isValidNamespace, "Operator.Namespace", operatorNamespace)
+
+		//Update status with Invalid Namespace
+		if err := r.updateStatusWithInvalidNamespace(reqLogger, request); err != nil {
+			return reconcile.Result{}, err
+		}
+
 		return reconcile.Result{}, nil
 	}
 
--- a/pkg/controller/mobilesecurityservice/status.go
+++ b/pkg/controller/mobilesecurityservice/status.go
@@ -45,6 +45,40 @@ func (r *ReconcileMobileSecurityService) updateStatus(reqLogger logr.Logger, con
 	return nil
 }
 
+// updateBindStatusWithInvalidNamespace returns error when status regards the all required resources could not be updated
+// DEPRECATED
+func (r *ReconcileMobileSecurityService) updateStatusWithInvalidNamespace(reqLogger logr.Logger, request reconcile.Request) error {
+	reqLogger.Info("Updating Bind App Status for the MobileSecurityServiceApp")
+
+	// Get the latest version of CR
+	instance, err := r.fetchInstance(reqLogger, request)
+	if err != nil {
+		return err
+	}
+
+	status := "Invalid Namespace"
+
+	//Update Bind CR Status with OK
+	if !reflect.DeepEqual(status, instance.Status.AppStatus) {
+		// Get the latest version of the CR in order to try to avoid errors when try to update the CR
+		instance, err := r.fetchInstance(reqLogger, request)
+		if err != nil {
+			return err
+		}
+
+		// Set the data
+		instance.Status.AppStatus = status
+
+		// Update the CR
+		err = r.client.Status().Update(context.TODO(), instance)
+		if err != nil {
+			reqLogger.Error(err, "Failed to update Status for the MobileSecurityService Bind")
+			return err
+		}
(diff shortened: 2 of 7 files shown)


camilamacedo86:
Hi @davidffrench,

I hope that the following explanation clarifies it better for you.  

Namespace scope means that the operator has access just to namespaced where it was "installed applied". When we say that the operator has cluster-scope it means that the operator has rules whcih allow his access to the cluster ( more than 1 namespace ). It is not about said that it works just in some specific namespace. 

May, if it not clarifies your understanding over it let's do a bluejeans. Also, could let me know why you blocked the PR? Are you not sure if we should do what you asked or did you find some issue with it?

davidffrench:
> Hi @davidffrench,
> 
> I hope that the following explanation clarifies it better for you.
> 
> The concept of namespaced means that the operator will NOT be able to work with more than 1 namespace. Which means that its rules are applied to allow access just in the current namespace. So it will not have RCBA roles to access objects applied in other ones but doesn't mean that you cannot create/install N times the same operator in many namespaces of the cluster.
> 
> May, if it not clarifies your understanding over it let's do a bluejeans. Also, could let me know why you blocked the PR? Are you not sure if we should do what you asked or did you find some issue with it?

My question is, what would have to a CR that is applied in a namespace that the operator is not watching?

It is blocked because it needs tests for the new functionality.

camilamacedo86:
Hi @davidffrench,

I think it has no relation with the impl at all. The question here is, have we the need to add the status == invalid namespace. If yes, probably we will never be able to solve it properly when the feature is released. 

IMHO, we should do nothing and has no need to add the status. Will someone project read this status? 
If not, if it is just to troubleshooting the logs are enough. 

Conclusion: I do not think that we should do it at all. It is not required.

davidffrench:
> Hi @davidffrench,
> 
> In POV it has no relation to the framework at all. The question here is:
> 
> * Have we the need to add the status == invalid namespace?
> * Will someone project read this status?
> 
> If yes, then it is horrible news since we will probably never able to solve it properly with the feature be released.
> 
> IMHO, we should do nothing and has no need to add the status. Regards troubleshooting the logs are enough in this scenario.

The problem I see is that the person applying the CR may not have access to look at the operator logs. Using pods as an example, I would hope there is some feedback telling me why a pod has not deployed correctly rather than just leaving the status blank and having to look through logs. The same applies here in my opinion, a user should not have to look through logs to understand why the creation of a CR has failed for any reason, this should be bubbled back up through the CR status.

camilamacedo86:
Hi @davidffrench,

> The problem I see is that the person applying the CR may not have access to look at the operator logs.  

In our context, the person never will apply the CR it is the job of another project and it will be impl in the MDC/RHMI. All is transparent for our users. 

The CR Status is for we be able to programatically check if all is ok or not. E.g Monitor Process.

camilamacedo86:
Hi @davidffrench,

> The problem I see is that the person applying the CR may not have access to look at the operator logs.  

In our context, the person never will apply the CR it is the job of another project and it will be impl in the MDC/RHMI. All is transparent for our users. 

The CR Status is for we be able to programatically check if all is ok or not. E.g Monitor Process.

davidffrench:
> Hi @davidffrench,
> 
> > The problem I see is that the person applying the CR may not have access to look at the operator logs.
> 
> My understanding so far is:
> 
> In our context, the person never will apply the CR it is the job of another project and it will be impl in the MDC/RHMI. All is transparent for our users. Actually, they will not have this info.
> 
> The CR Status is for our projects be able to programmatically check if all is ok or not. Then do not have the OK status there is enough to them in POV. E.g Monitor Process.

That is correct, currently the only thing that will be creating the CR will be MDC, so this is not a big concern right now. It would still be nice to implement for the future. And yes the status should give the current state of the CR for whatever party needs it.

camilamacedo86:
Hi @davidffrench,

You raised this need then shows that after doing the PR you are no longer sure if you would like or not this impl. IMHO, as explained before, we should not do it, however, please could you let me know your final position? 

If you would like to have this status in this scenario as you asked then we should verify and merge it, if not so we should close this PR. 

I am looking for your final position.

camilamacedo86:
Hi@davidffrench, 

Regards your comment `@camilamacedo86 Yes I still think this change is needed.`

So, could you let me know why this PR is blocked? Could you please let me know what change do you expected to be made here for we are able to merge it?

davidffrench:
@camilamacedo86 It was blocked due to no tests being included.

camilamacedo86:
Did you check it now @davidffrench ? Could we merge it? All are fine?

PR #102, opened 2019-05-24
AEROGEAR-8951 - Service app controller tests

damienomurchu:
## Motivation
AEROGEAR-8951 

## What
Add test coverage for app controller

## Why
Add test coverage

## How
Add test coverage for methods in app controller

## Verification Steps

Confirm tests complete successfully:
- confirm tests run locally (`GOCACHE=off go test -cover github.com/aerogear/mobile-security-service-operator/pkg/controller/mobilesecurityserviceapp -v`)
- confirm tests are passing in CI in this PR

Confirm changes have not affected or broken the operator:
- [ ] Install the operator with an image that you will build for this test then:
- [ ] Create an app and check if it was created in the service side
- [ ] Delete an app and check if it was deleted in the service side
- [ ] Update the app name in the APP CR and create it again and check if it was updated in the service side.

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/pkg/controller/mobilesecurityserviceapp/controller_test.go
+++ b/pkg/controller/mobilesecurityserviceapp/controller_test.go
@@ -1,0 +1,305 @@
+package mobilesecurityserviceapp
+
+import (
+	errs "errors"
+	mobilesecurityservicev1alpha1 "github.com/aerogear/mobile-security-service-operator/pkg/apis/mobilesecurityservice/v1alpha1"
+	"reflect"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/errors"
+
+	"github.com/aerogear/mobile-security-service-operator/pkg/models"
+
+	"github.com/aerogear/mobile-security-service-operator/pkg/service"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestReconcileMobileSecurityServiceApp_create(t *testing.T) {
+	type fields struct {
+		client client.Client
+		scheme *runtime.Scheme
+	}
+	type args struct {
+		instance   *mobilesecurityservicev1alpha1.MobileSecurityServiceApp
+		kind       string
+		serviceURL string
+		request reconcile.Request
+		err     error
+	}
+	tests := []struct {
+		name      string
+		fields    fields
+		args      args
+		want      reconcile.Result
+		wantErr   bool
+		wantPanic bool
+	}{
+		{
+			name: "should return a configmap",
+			fields: fields{
+				scheme: scheme.Scheme,
+			},
+			args: args{
+				kind:     ConfigMap,
+				err:      errors.NewInternalError(errs.New("Internal Server Error")),
+			},
+			want:    reconcile.Result{Requeue: true},
+			wantErr: false,
+		},
+		{
+			name: "should return an error when type other than CONFIGMAP specified",
+			fields: fields{
+				scheme: scheme.Scheme,
+			},
+			args: args{
+				kind:     "WRONG_KIND",
+				err:      errors.NewInternalError(errs.New("Internal Server Error")),
+			},
+			want:      reconcile.Result{},
+			wantErr:   true,
+			wantPanic: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			objs := []runtime.Object{
+				&instance,
+				&mssInstance,
+				&route,
+			}
+
+			r := buildReconcileWithFakeClientWithMocks(objs, t)
+
+			reqLogger := log.WithValues("Request.Namespace", &instance.Namespace, "Request.Name", &instance.Name)
+
+			// testing if the panic() function is executed
+			defer func() {
+				r := recover()
+				if (r != nil) != tt.wantPanic {
+					t.Errorf("ReconcileMobileSecurityService.buildFactory() recover = %v, wantPanic = %v", r, tt.wantPanic)
+				}
+			}()
+
+			err := r.create(&instance, tt.args.kind, tt.args.serviceURL, reqLogger, tt.args.request)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("ReconcileMobileSecurityServiceApp.create() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+		})
+	}
+}
+
+func TestReconcileMobileSecurityServiceApp_buildFactory(t *testing.T) {
+	type fields struct {
+		client client.Client
+		scheme *runtime.Scheme
+	}
+	type args struct {
+		instance   *mobilesecurityservicev1alpha1.MobileSecurityServiceApp
+		kind       string
+		serviceURL string
+	}
+	tests := []struct {
+		name      string
+		fields    fields
+		args      args
+		want      reflect.Type
+		wantErr   bool
+		wantPanic bool
+	}{
+		{
+			name: "should create a ConfigMap",
+			fields: fields{
+				scheme: scheme.Scheme,
+			},
+			want: reflect.TypeOf(&corev1.ConfigMap{}),
+			args: args{
+				kind:       ConfigMap,
+				serviceURL: "service-url",
+			},
+		},
+		{
+			name: "Should panic when trying to create unrecognized object type",
+			fields: fields{
+				scheme: scheme.Scheme,
+			},
+			args: args{
+				kind:       "UNDEFINED",
+				serviceURL: "service-url",
+			},
+			wantPanic: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			objs := []runtime.Object{
+				&instance,
+			}
+
+			r := buildReconcileWithFakeClientWithMocks(objs, t)
+
+			reqLogger := log.WithValues("Request.Namespace", &instance.Namespace, "Request.Name", &instance.Name)
+
+			// testing if the panic() function is executed
+			defer func() {
+				r := recover()
+				if (r != nil) != tt.wantPanic {
+					t.Errorf("ReconcileMobileSecurityServiceApp.buildFactory() recover = %v, wantPanic = %v", r, tt.wantPanic)
+				}
+			}()
+
+			configMap := r.buildFactory(reqLogger, &instance, tt.args.kind, tt.args.serviceURL)
+			if configMap == nil {
+				t.Errorf("ReconcileMobileSecurityServiceApp.buildFactory() - received no config type")
+				return
+			}
+
+			gotType := reflect.TypeOf(configMap)
+			expectedType := reflect.TypeOf(&corev1.ConfigMap{})
+			if gotType != expectedType {
+				t.Errorf("ReconcileMobileSecurityServiceApp.buildFactory() - received wrong config type")
+				return
+			}
+		})
+	}
+}
+
+func TestReconcileMobileSecurityServiceApp_Reconcile_FetchInstance(t *testing.T) {
+
+	objs := []runtime.Object{
+		&instanceInvalidName,
+		&mssInstance,
+		&route,
+	}
+
+	r := buildReconcileWithFakeClientWithMocks(objs, t)
+
+	// mock request to simulate Reconcile() being called on an event for a watched resource
+	req := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      instance.Name,
+			Namespace: instance.Namespace,
+		},
+	}
+
+	res, err := r.Reconcile(req)
+	// should return without error or requeueing request if App is not found
+	if res.Requeue == true || err != nil {
+		t.Fatalf("returned unexpectedly after attempting to fetch instance")
+	}
+}
+
+func TestReconcileMobileSecurityServiceApp_Reconcile_Deletion(t *testing.T) {
+	// required to test validity of namespace as namespace check won't be hit after
+	// fetch instance call beforehand in reconcile loop
+	objs := []runtime.Object{
+		&instanceForDeletion,
+		&mssInstance,
+		&route,
+	}
+
+	r := buildReconcileWithFakeClientWithMocks(objs, t)
+
+	// mock request to simulate Reconcile() being called on an event for a watched resource
+	req := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      instance.Name,
+			Namespace: instance.Namespace,
+		},
+	}
+
+	// mock fetchBindAppRestServiceByAppID http call
+	fetchBindAppRestServiceByAppID = func(serviceURL string, instance *mobilesecurityservicev1alpha1.MobileSecurityServiceApp, reqLogger logr.Logger) (*models.App, error) {
+		app := models.App{ID: "1234", AppName: "test", AppID: "test-app-id"}
+		return &app, nil
+	}
+
+	service.DeleteAppFromServiceByRestAPI = func(serviceAPI string, id string, reqLogger logr.Logger) error {
+		return nil
+	}
+
+	res, err := r.Reconcile(req)
+	// should return without error or requeueing request if App is deleted
+	if res.Requeue == true || err != nil {
+		t.Fatalf("returned unexpectedly after attempting to delete app")
+	}
+
+
+}
+
+func TestReconcileMobileSecurityServiceApp_Reconcile(t *testing.T) {
+	// objects to track in the fake client
+	objs := []runtime.Object{
+		&instance,
+		&mssInstance,
+		&route,
+	}
+
+	r := buildReconcileWithFakeClientWithMocks(objs, t)
+
+	// mock request to simulate Reconcile() being called on an event for a watched resource
+	req := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      instance.Name,
+			Namespace: instance.Namespace,
+		},
+	}
+
+	// mock fetchBindAppRestServiceByAppID http call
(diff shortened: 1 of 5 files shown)


pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+}
+
+func getReconciler(objs []runtime.Object) *ReconcileMobileSecurityServiceApp {
+	s := scheme.Scheme
+
+	routev1.AddToScheme(s)
+
+	s.AddKnownTypes(mobilesecurityservicev1alpha1.SchemeGroupVersion, &mobilesecurityservicev1alpha1.MobileSecurityServiceApp{})
+
+	s.AddKnownTypes(mobilesecurityservicev1alpha1.SchemeGroupVersion, &mobilesecurityservicev1alpha1.MobileSecurityService{})
```
camilamacedo86: Here is not required all this adds.

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+		t.Fatalf("update bind status: (%v)", err)
+	}
+
+	// res, err = r.Reconcile(req)
+}
+
+func getReconciler(objs []runtime.Object) *ReconcileMobileSecurityServiceApp {
+	s := scheme.Scheme
+
+	routev1.AddToScheme(s)
```
camilamacedo86: Could you do as it is not for the ReconcileMobileSecurityService?

camilamacedo86:
Hi @damienomurchu,

Very nice your PR. Just some small things. I'd suggest you check the tests on MSS. You can cut and paste the files and do small adjustments since they will be quite similar.  Note that it not covering the same. 

Screenshot 2019-05-24 at 16 07 39

pkg/controller/mobilesecurityserviceapp/controller_test.go
damienomurchu: yes, I need to refactor to use the same approach for the mocks as in ReconcileMobileSecurityService and that will be done in this PR :+1:

damienomurchu:
@camilamacedo86 yes, I have been focussing in the tests in the controller until now. I will be checking the other test files and see what can be re-used

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+			MemoryRequest:   "512Mi",
+			ClusterProtocol: "http",
+			ConfigMapName:   "mss-config",
+			Port:            1234,
+			RouteName:       "mss-route",
+		},
+	}
+
+	route = routev1.Route{
+		TypeMeta: v1.TypeMeta{
```
craicoverflow: TypeMeta struct can be completely removed as this is injected automatically.

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+			wantErr:   true,
+			wantPanic: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			objs := []runtime.Object{tt.args.instance}
+
+			tt.fields.scheme.AddKnownTypes(mobilesecurityservicev1alpha1.SchemeGroupVersion, tt.args.instance)
```
craicoverflow: This logic is already implemented in [`getReconciler`](https://github.com/aerogear/mobile-security-service-operator/pull/102/files#diff-4258282b1e19e790adfeea6e6c9c8db7R325).

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+			},
+			wantPanic: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			objs := []runtime.Object{tt.args.instance}
+
+			tt.fields.scheme.AddKnownTypes(mobilesecurityservicev1alpha1.SchemeGroupVersion, tt.args.instance)
```
craicoverflow: This logic is already implemented in [`getReconciler`](https://github.com/aerogear/mobile-security-service-operator/pull/102/files#diff-4258282b1e19e790adfeea6e6c9c8db7R325).

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -0,0 +1,340 @@
...
+			}
+		})
+	}
+}
+
+func TestReconcileMobileSecurityServiceApp_Reconcile(t *testing.T) {
+	// objects to track in the fake client
+	objs := []runtime.Object{
+		&instance,
+		&instanceTwo,
```
craicoverflow: I believe only `&instance` should be in this runtime object. The others are not created yet and should be created in the Reconcile method.

See https://github.com/aerogear/mobile-security-service-operator/blob/459c1e530253f6387fa4cc39f4a6c031b6a112fe/pkg/controller/mobilesecurityservicedb/controller_test.go#L228

damienomurchu: this should not have been there anyway, as the 2nd instance is not used in the reconcile method, so have removed it

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -264,7 +250,6 @@ func TestReconcileMobileSecurityServiceApp_Reconcile(t *testing.T) {
 	// objects to track in the fake client
 	objs := []runtime.Object{
 		&instance,
```
craicoverflow: I think the only item in `objs` should be `&instance`. Everything else is created during the reconcile loop.

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -223,18 +211,16 @@ func TestReconcileMobileSecurityServiceApp_buildFactory(t *testing.T) {
...
-			tt.fields.scheme.AddKnownTypes(mobilesecurityservicev1alpha1.SchemeGroupVersion, tt.args.instance)
-
-			cl := fake.NewFakeClient(objs...)
-
-			r := &ReconcileMobileSecurityServiceApp{
-				client: cl,
-				scheme: tt.fields.scheme,
+			objs := []runtime.Object{
+				&instance,
+				&instanceTwo,
```
craicoverflow: I think the only item in `objs` should be `&instance`. Everything else is created during the reconcile loop.

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -0,0 +1,325 @@
...
+	mobilesecurityservicev1alpha1 "github.com/aerogear/mobile-security-service-operator/pkg/apis/mobilesecurityservice/v1alpha1"
+	"github.com/aerogear/mobile-security-service-operator/pkg/utils"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	routev1 "github.com/openshift/api/route/v1"
+
+	corev1 "k8s.io/api/core/v1"
+	// "k8s.io/api/extensions/v1beta1"
```
craicoverflow: This can be removed.

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -0,0 +1,325 @@
...
+func TestReconcileMobileSecurityServiceApp_create(t *testing.T) {
+	type fields struct {
+		client client.Client
+		scheme *runtime.Scheme
+	}
+	type args struct {
+		instance   *mobilesecurityservicev1alpha1.MobileSecurityServiceApp
+		kind       string
+		serviceURL string
+		// reqLogger  logr.Logger
```
craicoverflow: This can be removed.

pkg/controller/mobilesecurityserviceapp/controller_test.go
craicoverflow: As @camilamacedo86 mentioned, many of these can be removed. Here is the same method but for `mobilesecurityservicedb`:

```go
// Creates a new reconciler with a fake client
func newReconcilerWithFakeClient(objs []runtime.Object) *ReconcileMobileSecurityServiceDB {
	s := scheme.Scheme

	s.AddKnownTypes(v1alpha1.SchemeGroupVersion, &v1alpha1.MobileSecurityServiceDB{})

	// create a fake client to mock API calls
	cl := fake.NewFakeClient(objs...)
	// create a ReconcileMobileSecurityService object with the scheme and fake client
	return &ReconcileMobileSecurityServiceDB{client: cl, scheme: s}
}

```

pkg/controller/mobilesecurityserviceapp/controller_test.go
damienomurchu: won't mssinstance & route be needed? app controller doesn't create those

pkg/controller/mobilesecurityserviceapp/controller_test.go
damienomurchu: same comment as above

pkg/controller/mobilesecurityserviceapp/controller_test.go
craicoverflow: You're right, my apologies!

damienomurchu:
@craicoverflow @camilamacedo86 all changes actioned. 

Can't really drill further down any of the tests in the controller at this point, and would suggest this PR gets merged, and test coverage can be improved by adding coverage for other files in the package to cover any gaps. 

@craicoverflow I was able to add more coverage to the controller file than when you last checked when it was 42%, fyi

camilamacedo86:
Hi @damienomurchu,

You can just copy and past the status and controller test from the mobilesecurityservice pkg, then small changes will be required. Note that since you add to test the coverage should be increased and not decreased.   

c/c @craicoverflow

pkg/controller/mobilesecurityserviceapp/fetches.go
```diff
@@ -30,6 +30,6 @@ func (r *ReconcileMobileSecurityServiceApp) fetchSDKConfigMap(reqLogger logr.Log
 }
 
 //fetchBindAppRestServiceByAppID return app struct from Mobile Security Service Project/REST API or error
-func fetchBindAppRestServiceByAppID(serviceURL string, instance *mobilesecurityservicev1alpha1.MobileSecurityServiceApp, reqLogger logr.Logger) (*models.App, error) {
+var fetchBindAppRestServiceByAppID = func(serviceURL string, instance *mobilesecurityservicev1alpha1.MobileSecurityServiceApp, reqLogger logr.Logger) (*models.App, error) {
```
camilamacedo86: Why it was changed?

pkg/service/client.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 //DeleteAppFromServiceByRestAPI delete the app object in the service
-func DeleteAppFromServiceByRestAPI(serviceAPI string, id string, reqLogger logr.Logger) error {
+var DeleteAppFromServiceByRestAPI = func(serviceAPI string, id string, reqLogger logr.Logger) error {
```
camilamacedo86: Why you change it? 
This are func...

pkg/controller/mobilesecurityserviceapp/fetches.go
craicoverflow: These are still functions and they should work the exact same. Assigning them as a variable like this means the variable can be reassigned in the tests.

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -0,0 +1,308 @@
...
+	req := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      instance.Name,
+			Namespace: instance.Namespace,
+		},
+	}
+
+	res, err := r.Reconcile(req)
+	// should return without error or requeueing request if App is not found
+	if (res.Requeue == true) || (err != nil) {
```
craicoverflow: Brackets are not needed here.

pkg/controller/mobilesecurityserviceapp/fakeclient_test.go
```diff
@@ -0,0 +1,31 @@
...
+	routev1 "github.com/openshift/api/route/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"testing"
+)
+
+//buildReconcileWithFakeClientWithMocks return reconcile with fake client, schemes and mock objects
+func buildReconcileWithFakeClientWithMocks(objs []runtime.Object, t *testing.T) *ReconcileMobileSecurityServiceApp {
```
craicoverflow: You don't need `WithMocks` in the function name, IMO :)

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -0,0 +1,308 @@
...
+		return &app, nil
+	}
+
+	service.DeleteAppFromServiceByRestAPI = func(serviceAPI string, id string, reqLogger logr.Logger) error {
+		return nil
+	}
+
+	res, err := r.Reconcile(req)
+	// should return without error or requeueing request if App is deleted
+	if (res.Requeue == true) || (err != nil) {
```
craicoverflow: Brackets are not needed in Go :)

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -0,0 +1,308 @@
...
+				serviceURL: "service-url",
+			},
+			wantPanic: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			objs := []runtime.Object{
+				&instance,
```
craicoverflow: Only &instance is required in `objs` for `buildFactory`.

pkg/controller/mobilesecurityserviceapp/fetches.go
camilamacedo86: Why create the var? 
I do not think that it should be made. The test can call the funcs.

craicoverflow:
> Revert the changes made to change to assign the result of the funcs to vars.

@damienomurchu did you verify that the operator functions normally with your assignment of those functions to variables?

pkg/controller/mobilesecurityserviceapp/fetches.go
damienomurchu: @camilamacedo86 its required to make it a var in order to mock the function for the controller tests. The function should work exactly as before - are you having issues?

pkg/controller/mobilesecurityserviceapp/fakeclient_test.go
damienomurchu: Just following the convention introduced by @camilamacedo86 in a previous refactor of the `mobileservice` pkg tests - are you ok if this is left as-is in this PR?

pkg/controller/mobilesecurityserviceapp/fetches.go
camilamacedo86: Hi @damienomurchu,
 
I do not think it still working as before because the project is calling funcs and not using vars.

damienomurchu:
> > Revert the changes made to change to assign the result of the funcs to vars.
> 
> @damienomurchu did you verify that the operator functions normally with your assignment of those functions to variables?

I haven't run any specific e2e verification of the operator functionality with these changes, I will touch base with @camilamacedo86 however and verify that the changes are ok.

I've made the changes you requested, except the function name which I left as-is to be consistent with what's in the `mobileservice` pkg

pkg/controller/mobilesecurityserviceapp/fetches.go
craicoverflow: Have either of you tested that it works? @damienomurchu\'s solution is also using functions.

https://golangbot.com/first-class-functions/

pkg/controller/mobilesecurityserviceapp/fakeclient_test.go
craicoverflow: Yes.

pkg/controller/mobilesecurityserviceapp/fetches.go
damienomurchu: > Nearly ready to approve. The only thing I think you should be covering in the Reconcile which you are not is [`hasMandatorySpecs`](https://github.com/aerogear/mobile-security-service-operator/blob/a30293abc535ff29c4434a3b473f8086378e80b3/pkg/controller/mobilesecurityserviceapp/controller.go#L181). Once that is covered I'm happy to approve

@craicoverflow I will try to add that test now - is it a problem if it is added in another PR though? (feel this PR has lingered a bit & would like to get it merged once @camilamacedo86 is happy to see it merged)

camilamacedo86: Hi @damienomurchu, 

Sorry, but I am not happy. I . think that this PR will break the master branch. 
Did you try to test the project after your change? note that since you create the vars it creates an impact in wll flow as create-app, delete-app and update-app.

craicoverflow: @damienomurchu yes it could be added in another PR. I'd like to see verification on @camilamacedo86`'s point.

damienomurchu: @craicoverflow I'm not sure it's possible to test the mandatory spec for a mobile service app as the spec for a mobile only [includes AppID and AppName](https://github.com/damienomurchu/mobile-security-service-operator/blob/master/deploy/crds/examples/mobile-security-service_v1alpha1_mobilesecurityserviceapp_cr.yaml#L6-L8), and either of those being invalid will fail the fetch instance check before the mandatory spec check

craicoverflow: Thanks for finding that out :) I'll dismiss my review.

craicoverflow:
> However, shows that something still wrong here. Why the coverage decreased when more tests were added? Is it missing a rebased? @craicoverflow could you help us with?

This is fine. @damienomurchu has only added tests for the controller file, whereas the entire package is now being counted in the coverage report. The other files in the package are bringing coverage down. @damienomurchu will be adding those in another PR :)

camilamacedo86:
So @craicoverflow I am approving it.

PR #103, opened 2019-05-24
feat(Allow usage of apply command for App CR) : Rebind without the need to delete and create again

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9268

## What
- Ensure that always will have just 1 ConfigMap created with the latest name of APP CR
- Allow usage of the apply command 
- Fix debug after all changes.
- Fix/improve tests after rebase
- Use the model from MSS instead of duplicate it.
- Upgrade operator-sdk from 0.7.0 to the latest version available 0.8.1 

## Verification Steps
Use the image:  docker.io/cmacedo/mobile-security-service-operator:AEROGEAR-9268

1. Install the oper
2. Create the app in the specific namespace
3. Check that a configMap was created for this app
4. Update the name into the Example APP CR 
5. Run the command `make create-app` again
6. It should have just one confgMap with the new name and not 2. 

## Checklist:

- [x] Changes have been successfully verified by another team member 

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

Screenshot 2019-05-28 at 10 45 44


scripts/export_local_envvars.sh
```diff
@@ -2,5 +2,6 @@
 
 export OPERATOR_NAME=mobile-security-service-operator
 export APP_NAMESPACES=mobile-security-service-apps
+export WATCH_NAMESPACE=mobile-security-service
 echo $OPERATOR_NAME
```
damienomurchu: remove the the `echo` statements?

.gitignore
```diff
@@ -6,6 +6,7 @@ vendor/
 build/_output
 build/_test
 cmd/manager/debug
+cmd/manager/main
```
damienomurchu: combine both with `cmd/manager/*` ?

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -62,6 +62,17 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 var _ reconcile.Reconciler = &ReconcileMobileSecurityServiceApp{}
 
+//Update the factory object and requeue
+func (r *ReconcileMobileSecurityServiceApp) delete(obj runtime.Object, reqLogger logr.Logger) error {
+	err := r.client.Delete(context.TODO(), obj)
+	if err != nil {
+		reqLogger.Error(err, "Failed to delete obj", "obj:", obj)
+		return err
+	}
+	reqLogger.Info("Deleted successfully")
```
damienomurchu: Should we need to include the object in the log to show the object deleted?

pkg/controller/mobilesecurityserviceapp/finalizer_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	mobilesecurityservicev1alpha1 "github.com/aerogear/mobile-security-service-operator/pkg/apis/mobilesecurityservice/v1alpha1"
+	"github.com/aerogear/mobile-security-service-operator/pkg/models"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestReconcileMobileSecurityServiceApp_addFinilizer(t *testing.T) {
```
damienomurchu:
```
func TestReconcileMobileSecurityServiceApp_addFinalizer(t *testing.T) {

```

pkg/controller/mobilesecurityserviceapp/finalizer_test.go
```diff
@@ -0,0 +1,130 @@
...
+			reqLogger := log.WithValues("Request.Namespace", tt.args.request.Namespace, "Request.Name", tt.args.request.Name)
+
+			if err := r.addFinalizer(reqLogger, tt.fields.instance, tt.args.request); (err != nil) != tt.wantErr {
+				t.Errorf("ReconcileMobileSecurityService.updateStatus() error = %v, wantErr %v", err, tt.wantErr)
+			}
+		})
+	}
+}
+
+func TestReconcileMobileSecurityServiceApp_removeFinilizer(t *testing.T) {
```
damienomurchu:
```suggestion
func TestReconcileMobileSecurityServiceApp_removeFinalizer(t *testing.T) {

```

pkg/controller/mobilesecurityservicedb/controller.go
```diff
@@ -86,10 +86,10 @@ type ReconcileMobileSecurityServiceDB struct {
 func (r *ReconcileMobileSecurityServiceDB) update(obj runtime.Object, reqLogger logr.Logger) error {
 	err := r.client.Update(context.TODO(), obj)
 	if err != nil {
-		reqLogger.Error(err, "Failed to update Spec")
+		reqLogger.Error(err, "Failed to update Object", "obj:", obj)
 		return err
 	}
-	reqLogger.Info("Spec updated - return and create")
+	reqLogger.Info("Object updated")
```
damienomurchu: Do we need to include the updated obj in the log?

camilamacedo86: nops .. just in the case of it fails

pkg/controller/mobilesecurityserviceapp/finalizer_test.go
camilamacedo86: Done 👍

pkg/controller/mobilesecurityserviceapp/finalizer_test.go
camilamacedo86: done 👍

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: I do not think so. It will be too much for the logs. The info is already printed above of it. Just in the case of its failures.

.gitignore
camilamacedo86: It is not possible. The main.go is in this dir.

scripts/export_local_envvars.sh
camilamacedo86: It is used in the run-local and debug-setup command. The echo is important to check/print when the command is executed the values added for validation.

laurafitzgerald:
@camilamacedo86 the use `feat(configmap)` in the commit message doesn't sound right to me. I believe this is to do with updating the app cr and not the config map? correct me if I'm wrong.

camilamacedo86:
@laurafitzgerald,

It is regards do not keep the old configMap in the namespace when the appName is updated.

scripts/export_local_envvars.sh
camilamacedo86: Done 👍

Makefile
```diff
@@ -62,6 +62,11 @@ create-app-ns:
 create-app:
 	kubectl create -f deploy/crds/examples/mobile-security-service_v1alpha1_mobilesecurityserviceapp_cr.yaml
 
+.PHONY: update-app
```
davidffrench: I believe the `apply` command will also work for creation if it doesn't exist so this could be just changed in the above `create-app` command rather than creating a new make command.

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -97,10 +97,10 @@ type ReconcileMobileSecurityService struct {
 func (r *ReconcileMobileSecurityService) update(obj runtime.Object, reqLogger logr.Logger) error {
 	err := r.client.Update(context.TODO(), obj)
 	if err != nil {
-		reqLogger.Error(err, "Failed to update Spec")
+		reqLogger.Error(err, "Failed to update Object", "obj:", obj)
```
davidffrench: Object is very non specific in this case. It would be good to clarify what the object is. e.g. `Failed to update <type> Object`. replacing type with whatever it is in this case, hardcoding is fine. As I don't have the context of this function, I am not sure what the Object type actually is.

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -97,10 +97,10 @@ type ReconcileMobileSecurityService struct {
...
 		return err
 	}
-	reqLogger.Info("Spec updated - return and create")
+	reqLogger.Info("Updated successfully")
```
davidffrench: Updated what successfully? When looking through logs, just an updated successfully won't be very helpful to look at.

pkg/controller/mobilesecurityserviceapp/controller.go
davidffrench: Same point as above, the logs should say what has been deleted.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -62,6 +62,17 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 var _ reconcile.Reconciler = &ReconcileMobileSecurityServiceApp{}
 
+//Update the factory object and requeue
+func (r *ReconcileMobileSecurityServiceApp) delete(obj runtime.Object, reqLogger logr.Logger) error {
+	err := r.client.Delete(context.TODO(), obj)
+	if err != nil {
+		reqLogger.Error(err, "Failed to delete obj", "obj:", obj)
```
davidffrench: Same point as above,`obj` is to vague, what type of object has failed to be deleted.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -155,7 +165,11 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
 
 	//Check if the APP CR was marked to be deleted
 	isAppMarkedToBeDeleted := instance.GetDeletionTimestamp() != nil
-	if isAppMarkedToBeDeleted && len(instance.GetFinalizers()) > 0 {
+
+	// If the Service instance was not found and/or is marked to be deleted
+	// OR
+	// if the APP CR was marked to be deleted
+	if (isAppMarkedToBeDeleted && len(instance.GetFinalizers()) > 0) || mssInstance == nil || mssInstance.GetDeletionTimestamp() != nil {
```
davidffrench: @camilamacedo86 Since the mssInstance is initalised as `&mobilesecurityservicev1alpha1.MobileSecurityService{}` https://github.com/aerogear/mobile-security-service-operator/pull/103/files#diff-f0c3b61d21a85c4c894842a29ca22af2R151 . Is there a possibility that the mssInstance will ever be nil? And if there is, in what scenario will it be nil?

The way I read this logic is that if the mssInstance is nil and the app CR is not marked for deletion, it will still be deleted. This could be very dangerous if the mssInstance is returned as nil in a case where we do not want to delete the app CR.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -221,13 +235,29 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
+	reqLogger.Info("Checking if the configMap already exists ...")
+	// Check if ConfigMap for the app exist, if not create one.
+	if _, err := r.fetchConfigMap(reqLogger, instance); err != nil {
 		if err := r.create(instance, ConfigMap, publicServiceURLAPI, reqLogger, request); err != nil {
 			return reconcile.Result{}, err
 		}
 	}
 
+	reqLogger.Info("Checking if a configMap for the same appId already exists with another name ...")
+	// Check if has already a confgMap for the same appId, if yes and the name is not the same then remove it
```
davidffrench: In a scenario where the appName has changed, this logic is deleting the old one, why not just update the existing ConfigMap?

pkg/controller/mobilesecurityservicedb/controller.go
```diff
@@ -86,10 +86,10 @@ type ReconcileMobileSecurityServiceDB struct {
 func (r *ReconcileMobileSecurityServiceDB) update(obj runtime.Object, reqLogger logr.Logger) error {
 	err := r.client.Update(context.TODO(), obj)
 	if err != nil {
-		reqLogger.Error(err, "Failed to update Spec")
+		reqLogger.Error(err, "Failed to update Object", "obj:", obj)
```
davidffrench: Same comment about the logger as in other controllers

pkg/controller/mobilesecurityservicedb/controller.go
davidffrench: Same comment about the logger as in other controllers

Makefile
camilamacedo86: Ok.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: Reasons:
- The update does no ensure all possible scenarios 
- The delete will work fine indeed with multi-thread events 
- The current impl will ensure that it will not have any configMap for the appID with an old name

pkg/controller/mobilesecurityservicedb/controller.go
camilamacedo86: - [ ] Note that it will log the object which means all its details indeed the kind.
- [ ] See that it is a generic impl / func that will update any runtime.object.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: - Note that it will log the object which means all its details indeed the kind.
- See that it is a generic impl / func that will update any runtime.object.
- Also, it is important to highlight that if we are tracing it in the logs above of it will find other logs with more details as well.

PS.: I changed to output the object in the case of success as well. It is important to highlight that it is as it was before.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: - Note that it will log the object which means all its details indeed the kind. It an object. 
- See that it is a generic impl / func that will update any runtime.object.
- Also, it is important to highlight that if we are tracing it in the logs above of it will find other logs with more details as well.

PS.: I changed to output the object in the case of success as well. It is important to highlight that it is as it was before.

pkg/controller/mobilesecurityservice/controller.go
camilamacedo86: - Note that it will log the object which means all its details indeed the kind. It an object. 
- See that it is a generic impl / func that will update any runtime.object.
- Also, it is important to highlight that if we are tracing it in the logs above of it will find other logs with more details as well.

PS.: I changed to output the object in the case of success as well. It is important to highlight that it is as it was before.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: The MSS will just be nill when the Service CR/CRD was deleted from the cluster.

pkg/controller/mobilesecurityservicedb/controller.go
camilamacedo86: - Note that it will log the object which means all its details indeed the kind.
- See that it is a generic impl / func that will update any runtime.object.
- Also, it is important to highlight that if we are tracing it in the logs above of it will find other logs with more details as well.

PS.: I changed to output the object in the case of success as well. It is important to highlight that it is as it was before.

pkg/controller/mobilesecurityservice/controller.go
camilamacedo86: I changed to output the object in the case of success as well

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: Done.

laurafitzgerald:
@camilamacedo86 I've completed the verification steps and the changes are working as expected based on the second last commit. I see there are some feedback items and the build is failing on a new commit now so I haven't added an approve yet. I added a verified label. Feel free to ping me when you want a recheck.

pkg/controller/mobilesecurityservice/controller.go
davidffrench: Perfect

pkg/controller/mobilesecurityserviceapp/controller.go
davidffrench: This is the only scenario where it can be nil?

pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -272,34 +272,191 @@ func TestReconcileMobileSecurityServiceApp_Reconcile(t *testing.T) {
...
+		t.Fatalf("reconcile: (%v)", err)
 	}
 
+	// Check the result of reconciliation to make sure it has the desired state
+	if res.Requeue {
+		t.Error("reconcile Requeue unexpected")
+	}
 }
 
+func TestReconcileMobileSecurityServiceApp_Reconcile_MMSDeleteCR(t *testing.T) {
```
damienomurchu:
```suggestion
func TestReconcileMobileSecurityServiceApp_Reconcile_MSSDeleteCR(t *testing.T) {
```

pkg/controller/mobilesecurityserviceapp/controller.go
davidffrench: Thanks @camilamacedo86 . This should be fine, I can't think of a scenario where deleting and recreating would be an issue unless something else is watching for event changes on that configmap but I don't believe there is.

pkg/controller/mobilesecurityserviceapp/controller_test.go
camilamacedo86: 👍

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: Yes. I cannot see another scenario.

PR #107, opened 2019-05-29
refac(controller) : refac to make clean the code and more performative

camilamacedo86:
## Motivation
feedback from a red hatter architecture which has been using this project as an example. 

## What
- Small improvements
- Fix some small issues that pass in the unit tests

## Verification Steps
Use the image docker.io/cmacedo/mobile-security-service-operator:IMP_FETCH and heck if the oper will be installed with success created/delete app

- `make create-all`
- `oc project mobile-security-service-apps`
- make create-app
- make delete-app

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/pkg/controller/mobilesecurityserviceapp/controller.go
+++ b/pkg/controller/mobilesecurityserviceapp/controller.go
@@ -10,7 +10,6 @@ import (
 	"github.com/go-logr/logr"
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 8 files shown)


pkg/controller/mobilesecurityserviceapp/controller_test.go
```diff
@@ -459,39 +467,4 @@ func TestReconcileMobileSecurityServiceApp_Reconcile_UpdateName(t *testing.T) {
 	if res.Requeue {
 		t.Error("reconcile Requeue unexpected")
 	}
-}
-
-func TestReconcileMobileSecurityServiceApp_Reconcile_InvalidAppNamespace(t *testing.T) {
```
camilamacedo86: It is not possible to be checked. 
When the project be running outside of the cluster (unit-tests or local env to debug-test it without the need to a new image) it will be returned as valid always.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -204,7 +185,7 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
 		}
 
 		// Check if the finalizer criteria is met and remove finalizer from the CR
-		if err := r.removeFinalizer(serviceAPI, reqLogger, request); err != nil {
+		if err := r.handleFinalizer(serviceAPI, reqLogger, request); err != nil {
```
austincunningham: removeFinializer makes more scene handle make be think of http requests

camilamacedo86: But it will make http requests to check first if it was removed from the service side them call the removeFinalizer which shows that in this way make more sense the change :-)

damienomurchu:
Just verified this according to your verification steps

camilamacedo86:
@damienomurchu missing the approval ;-)

PR #110, opened 2019-05-31
feat(PullIfNotPresent) : Apply the policy PullIfNotPresent for service images

camilamacedo86:
Closed because of the rebase issues. 
New PR was made for it : https://github.com/aerogear/mobile-security-service-operator/pull/111


pkg/controller/mobilesecurityservice/deployments.go
```diff
@@ -73,7 +74,7 @@ func buildApplicationContainer(m *mobilesecurityservicev1alpha1.MobileSecuritySe
 	return corev1.Container{
 		Image:           m.Spec.Image,
 		Name:            m.Spec.ContainerName,
-		ImagePullPolicy: corev1.PullAlways,
+		ImagePullPolicy: corev1.PullIfNotPresent,
```
austincunningham: Don't know if I like this change. Does it mean it won't pull a fresh image if the upstream image changes? would prefer PullAlways behavior

pkg/controller/mobilesecurityservice/deployments.go
```diff
@@ -55,8 +55,9 @@ func getDeploymentContainers(m *mobilesecurityservicev1alpha1.MobileSecurityServ
 
 func buildOAuthContainer(m *mobilesecurityservicev1alpha1.MobileSecurityService) corev1.Container {
 	return corev1.Container{
-		Image: "docker.io/openshift/oauth-proxy:v1.1.0",
-		Name:  "oauth-proxy",
+		Image:           "docker.io/openshift/oauth-proxy:v1.1.0",
```
wei-lee: would prefer this to be configurable via an env var. It is possible we need to switch to use quay.io for production environments.

pkg/controller/mobilesecurityservice/deployments.go
```diff
@@ -73,7 +74,7 @@ func buildApplicationContainer(m *mobilesecurityservicev1alpha1.MobileSecuritySe
 	return corev1.Container{
 		Image:           m.Spec.Image,
```
wei-lee: why do we allow the image to be specified in the spec? What if user specify a completely different image?

camilamacedo86: The user cannot specify an image because the operator is cluster scope which means that just the admin of the cluster has access to it. Is it make sense?

pkg/controller/mobilesecurityservice/deployments.go
camilamacedo86: IMHO, so we need to add it in the spec. If we would like to change we just update the specs of the CR which shows the right place to do that.

pkg/controller/mobilesecurityservice/deployments.go
camilamacedo86: Hi @austincunningham,

I understand that the PullIfNotPresent will always pull the latest tag if it is not present.
Please, let me know if I am wrong here @wei-lee?

wei-lee: Both of them will have use cases. Let's pick a default value (default could be empty and let openshift pick - https://docs.openshift.com/enterprise/3.0/architecture/core_concepts/builds_and_image_streams.html#image-pull-policy) and allow users to override in the spec.

pkg/controller/mobilesecurityservice/deployments.go
wei-lee: Spec is fine too, but it should definitely be configurable. 

Personally I don't think it is best practice to allow images to be specified via CRD spec: they rarely changes and allow them to be set in spec without some sort of validation could easily cause issues.  But also there isn't any recommendation about what is the best practice here, simply because the whole operator stuff is relatively new. So I guess we will wait and see.

pkg/controller/mobilesecurityservice/deployments.go
camilamacedo86: Tks @wei-lee I will do it.

pkg/controller/mobilesecurityservice/deployments.go
wei-lee: In here `user` I mean whoever wants to install the service, so it could be a cluster admin. My comment is really about we should try to minimise the possibility that someone could make a mistake, by providing less options in the spec, especially if they don't need to provide. If we have a default value for it then it fine too.

pkg/controller/mobilesecurityservice/deployments.go
camilamacedo86: Done :-)

pkg/controller/mobilesecurityservice/deployments.go
camilamacedo86: Done as requested.

pkg/controller/mobilesecurityservice/deployments.go
camilamacedo86: Closing since more interactions were made here and it did not receive updates which lead me to understand that are ok over it.

pkg/controller/mobilesecurityservice/deployments.go
austincunningham: seems to be a lot of files added to the pr. may need a rebase

davidffrench:
@camilamacedo86 As Austin mentioned previously, looks like something went wrong with the rebase. Please let me know once this is resolved, I would like to review this PR.

camilamacedo86:
@davidffrench @austincunningham fixed with the rebase.

camilamacedo86:
Hi, @wei-lee 

I added the impl with the default/constants values as you ask. Since we will use this approach it was made for all mandatory specs. In this way, if the CR (Service and/or DB) be applied without any spec it should work with the default values.

PR #111, opened 2019-06-01
Last Changes before first tag release

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9316

## What
- Allow cluster admin configure image policies (Optional)
- The default value will be empty as requested which will allow the OCP to decide. 
- Remove outdated info from README
- Reorganize CRs file and improve the comments to make clear what is each spec. 
- Add mandatory specs by constants as requested by @wei-lee 
- Do refectories regards comments and instance names (PS.: it is our last chance before since we do not create a tag yet :-))
- Add test to check mandatory specs

**NOTE:** The changes are spit in each commit. You can check the commits' changes to make easier the verification.  

## Why
- In this, the policy can be changed according to the business by its setup/config files. 

## Steps to verify
- Use the image: `docker.io/cmacedo/mobile-security-service-operator:changes `
- Install the operator and check if all still working
- Check the create/delete app and config Map
- Check that the db still using the configMap values for its EnvVars

IMPORTANT: The project is now fully covered by tests and if something brokes is unlike we do not get it so long in the CI. Also, we added a test to ensure the new feat to allowing the usage without specs.


pkg/controller/mobilesecurityservice/mandatory_specs.go
```diff
@@ -0,0 +1,137 @@
+package mobilesecurityservice
+
+import (
+	mobilesecurityservicev1alpha1 "github.com/aerogear/mobile-security-service-operator/pkg/apis/mobilesecurityservice/v1alpha1"
+)
+
+const (
```
camilamacedo86: @wei-lee now we can apply it without any spec and the "default" values will be applied automatically

pkg/controller/mobilesecurityservicedb/controller.go
```diff
@@ -92,31 +89,34 @@ func (r *ReconcileMobileSecurityServiceDB) update(obj runtime.Object, reqLogger
...
+func (r *ReconcileMobileSecurityServiceDB) buildFactory(db *mobilesecurityservicev1alpha1.MobileSecurityServiceDB, kind string, reqLogger logr.Logger) runtime.Object {
+	reqLogger.Info("Check "+kind, "into the namespace", db.Namespace)
 	switch kind {
 	case PVC:
-		return r.buildPVCForDB(instance)
+		return r.buildPVCForDB(db)
 	case Deployment:
-		return r.buildDBDeployment(instance, serviceInstance)
+		// FIXME: It should not be made here
+		// FIXME: The Service should find the db deployment and update its ENV VARs instead of we looking for it here for we are able to extract it and use it with any service
```
davidffrench: Will this be fixed in this PR?

pkg/controller/mobilesecurityservicedb/mandatory_specs.go
```diff
@@ -0,0 +1,97 @@
...
+	databaseMemoryRequest  = "512Mi"
+	databaseStorageRequest = "1Gi"
+)
+
+// addMandatorySpecsDefinitions will add the specs which are mandatory for Mobile Security Service CR in the case them
+// not be applied
+func addMandatorySpecsDefinitions(db *mobilesecurityservicev1alpha1.MobileSecurityServiceDB) {
+
+	/*
+		#   CR DB Resource
```
davidffrench: Small spacing issue here

pkg/controller/mobilesecurityservicedb/mandatory_specs.go
```diff
@@ -0,0 +1,97 @@
...
+		#   ---------------------
+	*/
+
+	if db.Spec.Size == 0 {
+		db.Spec.Size = size
+	}
+
+	/*
+		#	Environment Variables
```
davidffrench: Small spacing issue here

pkg/controller/mobilesecurityservicedb/controller.go
camilamacedo86: Not required be done now at all. I update the msg to make it clear

```go
	// If we decide to export the DB to be reused it for other operators (E.g create an operator for the PostgreSQL) then it should not be made here
	// The Service should find the DB deployment and update its ENV VARs instead of we looking for it here for we are able to extract it and use it with any service

```

pkg/controller/mobilesecurityservicedb/mandatory_specs.go
camilamacedo86: David, it is how the make fmt align it.

davidffrench:
Thanks @camilamacedo86 . IF I get time today, I will verify this PR but it is likely another member of the team will get to it before I do

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
```diff
@@ -4,32 +4,53 @@ kind: MobileSecurityService
 metadata:
   name: mobile-security-service
 spec:
+
+  # NOTE: If no specs be defined here the constants values in the code implementation will be applied
```
laurafitzgerald: Suggest: # NOTE: Default values will be applied by the operator if values are not specified below.

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
```diff
@@ -4,32 +4,53 @@ kind: MobileSecurityService
...
+  memoryLimit: "512Mi"
+  memoryRequest: "512Mi"
+  # Use the following spec if you would like to define the image policy
+  # containerImagePullPolicy: "IfNotPresent"
+
+  # ---------------------------------
+  # OAuth Proxy
+  # ---------------------------------
+
+  oAuthPort: 4180
```
laurafitzgerald: Do we need to expose the oauth values?

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservicedb_cr.yaml
```diff
@@ -4,22 +4,41 @@ kind: MobileSecurityServiceDB
 metadata:
   name: mobile-security-service-db
 spec:
+
+  # NOTE: If no specs be defined here the constants values in the code implementation will be applied
```
laurafitzgerald: Same suggestion as above.

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservicedb_cr.yaml
```diff
@@ -4,22 +4,41 @@ kind: MobileSecurityServiceDB
...
+
+  # ---------------------------------
+  # Database Container
+  # ---------------------------------
+
   # The imaged used in this project is from Red Hat. See more in https://docs.okd.io/latest/using_images/db_images/postgresql.html
   image: "centos/postgresql-96-centos7"
   containerName: "database"
-  # Following are the args which will be used as the key label for the environment variable of the database image.
+  # Use the following spec if you would like to define the image policy
```
laurafitzgerald: image policy -> image pull policy

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservicedb_cr.yaml
```diff
@@ -4,22 +4,41 @@ kind: MobileSecurityServiceDB
...
-  # Following the default values for the Env Variables which will be created by the operator
-  # for the Mobile Security Service Database deployment.
-  # This values will be used if the ConfigMap created and used by the Mobile Security Service be not found
+
+  # ---------------------------------
+  # Environment Variables
+  # ---------------------------------
+
+  # IMPORTANT: By default the database will use the values defined in the Mobile Security Service and shared via a Config Map.
+  # The following specs are just used just if the database be installed without the Service (standalone)
```
laurafitzgerald: Suggestion: The following specs are used if the .....

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -104,36 +104,36 @@ func (r *ReconcileMobileSecurityService) update(obj runtime.Object, reqLogger lo
 	return nil
 }
 
-//Create the factory object and requeue
-func (r *ReconcileMobileSecurityService) create(instance *mobilesecurityservicev1alpha1.MobileSecurityService, reqLogger logr.Logger, kind string) error {
-	obj := r.buildFactory(reqLogger, instance, kind)
-	reqLogger.Info("Creating a new ", "kind", kind, "Namespace", instance.Namespace)
+// Create the factory object and requeue
+func (r *ReconcileMobileSecurityService) create(service *mobilesecurityservicev1alpha1.MobileSecurityService, reqLogger logr.Logger, kind string) error {
```
laurafitzgerald: In some places the instance of the MobileSecurityService is named `service` and in others it's named `instance`. We should keep it consistent and in my opinion we should avoid both of these terms. `service` because there is a Service resource in openshift so this could be confusing and `instance` because that doesn't describe what's there, we could call it mobilesecurityservice or mss so that we know that it is an instance of the MobileSecurityService.

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -147,54 +147,51 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling Mobile Security Service ...")
 
-	//Fetch the MobileSecurityService instance
-	instance := &mobilesecurityservicev1alpha1.MobileSecurityService{}
-	instance, err := r.fetchInstance(reqLogger, request)
+	// Fetch the MobileSecurityService service
+	service := &mobilesecurityservicev1alpha1.MobileSecurityService{}
+	service, err := r.fetchServiceInstance(reqLogger, request)
```
laurafitzgerald: As with the above can we avoid using the work `Service` in the naming of this method as there is a `Service` resource already in Openshift. We're not fetching a Service here it's fetching the MobileSecurityService instance.

pkg/controller/mobilesecurityservice/deployments.go
```diff
@@ -11,14 +11,14 @@ import (
 )
 
 //buildDeployment returns the Deployment object using as image the MobileSecurityService App ( UI + REST API)
-func (r *ReconcileMobileSecurityService) buildDeployment(m *mobilesecurityservicev1alpha1.MobileSecurityService) *v1beta1.Deployment {
+func (r *ReconcileMobileSecurityService) buildDeployment(service *mobilesecurityservicev1alpha1.MobileSecurityService) *v1beta1.Deployment {
```
laurafitzgerald: same here regarding the naming of the object as above

pkg/controller/mobilesecurityservice/serviceaccount.go
```diff
@@ -9,18 +9,18 @@ import (
...
+			Name:        service.Name,
+			Namespace:   service.Namespace,
+			Annotations: buildOauthAnnotationWithRoute(service.Spec.RouteName),
 		},
 	}
 }
 
-func getAnnotation(route string) map[string]string {
+// buildOauthAnnotationWithRoute return required annotations for the Oauth setup
+func buildOauthAnnotationWithRoute(route string) map[string]string {
```
laurafitzgerald: 👍

pkg/controller/mobilesecurityserviceapp/configmaps.go
```diff
@@ -8,16 +8,16 @@ import (
 )
 
 // Returns the ConfigMap with the properties used to setup/config the Mobile Security Service Project
-func (r *ReconcileMobileSecurityServiceApp) buildAppSDKConfigMap(m *mobilesecurityservicev1alpha1.MobileSecurityServiceApp, serviceURL string) *corev1.ConfigMap {
+func (r *ReconcileMobileSecurityServiceApp) buildAppSDKConfigMap(app *mobilesecurityservicev1alpha1.MobileSecurityServiceApp, serviceURL string) *corev1.ConfigMap {
```
laurafitzgerald: 👍

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -136,7 +136,7 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
 	}
 
 	reqLogger.Info("Valid namespace for MobileSecurityServiceApp", "Namespace", request.Namespace)
-	reqLogger.Info("Checking for service instance ...")
+	reqLogger.Info("Checking for service app ...")
```
laurafitzgerald: Suggestion: Checking for app ...

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -167,21 +167,21 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
 			return reconcile.Result{}, nil
 		}
 
-		// If the CR was marked to be deleted before it finalizes the app need to be deleted from the Service
-		// Do request to get the app.ID to delete app
-		app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger)
+		// If the CR was marked to be deleted before it finalizes the appService need to be deleted from the Service
```
laurafitzgerald: again the use of `Service` here. I think we should be careful about how we use it. this doesn't sound right. `The appService needs to be deleted from the Service`?

pkg/controller/mobilesecurityservicedb/controller.go
```diff
@@ -92,31 +89,34 @@ func (r *ReconcileMobileSecurityServiceDB) update(obj runtime.Object, reqLogger
 	return nil
 }
 
-//Create the object and reconcile it
-func (r *ReconcileMobileSecurityServiceDB) create(instance *mobilesecurityservicev1alpha1.MobileSecurityServiceDB, serviceInstance *mobilesecurityservicev1alpha1.MobileSecurityService, kind string, reqLogger logr.Logger) error {
-	obj := r.buildFactory(instance, serviceInstance, kind, reqLogger)
-	reqLogger.Info("Creating a new ", "kind", kind, "Namespace", instance.Namespace)
+// Create the object and reconcile it
+func (r *ReconcileMobileSecurityServiceDB) create(db *mobilesecurityservicev1alpha1.MobileSecurityServiceDB, kind string, reqLogger logr.Logger) error {
```
laurafitzgerald: 👍

pkg/controller/mobilesecurityservicedb/helpers.go
```diff
@@ -13,14 +14,14 @@ func getDBLabels(name string) map[string]string {
 	return map[string]string{"app": "mobilesecurityservicedb", "mobilesecurityservicedb_cr": name, "name": "mobilesecurityservicedb"}
 }
 
-func (r *ReconcileMobileSecurityServiceDB) getDatabaseNameEnvVar(m *mobilesecurityservicev1alpha1.MobileSecurityServiceDB, serviceInstance *mobilesecurityservicev1alpha1.MobileSecurityService) corev1.EnvVar {
-	if r.hasAppConfigMap(m, serviceInstance) {
+func (r *ReconcileMobileSecurityServiceDB) getDatabaseNameEnvVar(db *mobilesecurityservicev1alpha1.MobileSecurityServiceDB, serviceConfigMapName string) corev1.EnvVar {
```
laurafitzgerald: Do we have to pass the whole db object here. Could we pass the attributes from that which are used? It looks like there are only two.

pkg/controller/mobilesecurityservicedb/helpers.go
```diff
@@ -29,35 +30,51 @@ func (r *ReconcileMobileSecurityServiceDB) getDatabaseNameEnvVar(m *mobilesecuri
...
+		Value: db.Spec.DatabaseName,
 	}
 }
 
-//Check if has App Config Map created
-func (r *ReconcileMobileSecurityServiceDB) hasAppConfigMap(m *mobilesecurityservicev1alpha1.MobileSecurityServiceDB, serviceInstance *mobilesecurityservicev1alpha1.MobileSecurityService) bool {
-	//if has not service instance return false
-	if len(serviceInstance.Name) < 1 {
-		return false
+func (r *ReconcileMobileSecurityServiceDB) getAppConfigMapName(db *mobilesecurityservicev1alpha1.MobileSecurityServiceDB) string {
```
laurafitzgerald: This function seems a bit nested. `getAppConfigMapName` calls `fetchServiceConfigMap`. I could be missing some context but maybe there's some improvement that could be made here?

pkg/controller/mobilesecurityservicedb/status.go
```diff
@@ -90,7 +90,7 @@ func (r *ReconcileMobileSecurityServiceDB) updateDeploymentStatus(reqLogger logr
 func (r *ReconcileMobileSecurityServiceDB) updateServiceStatus(reqLogger logr.Logger, request reconcile.Request) (*corev1.Service, error) {
 	reqLogger.Info("Updating Service Status for the MobileSecurityServiceDB")
 	// Get the latest version of the instance CR
-	instance, err := r.fetchInstance(reqLogger, request)
+	instance, err := r.fetchDBInstance(reqLogger, request)
```
laurafitzgerald: The instance is being fetched twice in this method and many others.

pkg/controller/mobilesecurityservice/controller.go
camilamacedo86: Hi @laurafitzgerald,

If you check the change is replace the instance or m .. for a name that makes clear what kind of instance it is. So, now it shows consistently. I will check if has someplace using instance yet.

pkg/controller/mobilesecurityservice/controller.go
camilamacedo86: I will replace in all for another name which can make clear.

pkg/controller/mobilesecurityservicedb/status.go
camilamacedo86: @laurafitzgerald I think that I already explain this before. 
It is required to avoid issues regards conflicts.

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
camilamacedo86: It is an option.

pkg/controller/mobilesecurityservice/controller.go
camilamacedo86: Done

pkg/controller/mobilesecurityservice/deployments.go
camilamacedo86: Done.

pkg/controller/mobilesecurityserviceapp/controller.go
camilamacedo86: Yes. I added the info `Service side`, however, IMHO it is quite clear the mean of the world since the context.

camilamacedo86: I will consider it solved since few changes were made.

pkg/controller/mobilesecurityservicedb/helpers.go
camilamacedo86: It is a better practice to pass the object instead of having many parameters in the func. Note that indeed we could want to use more attributes of this obj in the future.

pkg/controller/mobilesecurityservicedb/helpers.go
camilamacedo86: I improved the name but the funcs will work separately.

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
laurafitzgerald: Have we tested what would happen if the user provides other values? I don't believe we want to allow the user to provide a different image for the oauth proxy. I think this will break it.

pkg/controller/mobilesecurityservice/controller.go
```diff
@@ -147,54 +147,51 @@ func (r *ReconcileMobileSecurityService) Reconcile(request reconcile.Request) (r
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling Mobile Security Service ...")
 
-	//Fetch the MobileSecurityService instance
-	instance := &mobilesecurityservicev1alpha1.MobileSecurityService{}
-	instance, err := r.fetchInstance(reqLogger, request)
+	// Fetch the MobileSecurityService service
+	service := &mobilesecurityservicev1alpha1.MobileSecurityService{}
```
laurafitzgerald: Can we also use mss here instead of service so that we are consistent

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -167,21 +167,21 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
 			return reconcile.Result{}, nil
 		}
 
-		// If the CR was marked to be deleted before it finalizes the app need to be deleted from the Service
-		// Do request to get the app.ID to delete app
-		app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger)
+		// If the CR was marked to be deleted before it finalizes the appService need to be deleted from the Service Side
```
laurafitzgerald: I think it's fine to use app here. the request gets an app object as apposed to an appService object from the MSS.

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -167,21 +167,21 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
 			return reconcile.Result{}, err
 		}
 
-		// If the request works with success and the app was found then
+		// If the request works with success and the appService was found then
 		// Do request to delete it from the service
 		if app.ID != "" {
 			if err := service.DeleteAppFromServiceByRestAPI(serviceAPI, app.ID, reqLogger); err != nil {
-				reqLogger.Error(err, "Unable to delete app from Service", "App.ID", app.ID)
+				reqLogger.Error(err, "Unable to delete appService from Service", "App.ID", app.ID)
```
laurafitzgerald: app

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -167,21 +167,21 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
 				return reconcile.Result{}, err
 			}
-			reqLogger.Info("Successfully delete app ...")
+			reqLogger.Info("Successfully delete appService ...")
```
laurafitzgerald: app

pkg/controller/mobilesecurityserviceapp/controller.go
```diff
@@ -167,21 +167,21 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
...
-		app, err := fetchBindAppRestServiceByAppID(serviceAPI, instance, reqLogger)
+		// If the CR was marked to be deleted before it finalizes the appService need to be deleted from the Service Side
+		// Do request to get the appService.ID to delete appService
+		app, err := fetchBindAppRestServiceByAppID(serviceAPI, mssApp, reqLogger)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
 
-		// If the request works with success and the app was found then
+		// If the request works with success and the appService was found then
```
laurafitzgerald: app

pkg/controller/mobilesecurityservice/controller.go
camilamacedo86: Sure I will check it

pkg/controller/mobilesecurityservice/helpers.go
```diff
@@ -37,26 +36,26 @@ func buildAppEnvVars(m *mobilesecurityservicev1alpha1.MobileSecurityService) *[]
...
 
 // getOAuthArgsMap is a helper to get the []string with values required/used to set OAuth for the Mobile Security Service Project
-func getOAuthArgsMap(m *mobilesecurityservicev1alpha1.MobileSecurityService) []string {
+func getOAuthArgsMap(mss *mobilesecurityservicev1alpha1.MobileSecurityService) []string {
 	return []string{
 		"--http-address=0.0.0.0:4180",
 		"--https-address=",
 		"--provider=openshift",
-		fmt.Sprintf("--openshift-service-account=%s", m.Name),
+		fmt.Sprintf("--openshift-mss-account=%s", mss.Name),
```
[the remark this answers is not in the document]
camilamacedo86: Sure thanks.

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
camilamacedo86: So, it should be fixed where it is used. I will change that.

camilamacedo86:
@laurafitzgerald all changes addressed. Also, a new image was built as well. Please feel free to check it.

pkg/controller/mobilesecurityservice/services.go
```diff
@@ -31,30 +31,30 @@ func (r *ReconcileMobileSecurityService) buildProxyService(m *mobilesecurityserv
...
-	controllerutil.SetControllerReference(m, ser, r.scheme)
+	// Set MobileSecurityService mss as the owner and controller
+	controllerutil.SetControllerReference(mss, ser, r.scheme)
 	return ser
 }
 
-func (r *ReconcileMobileSecurityService) buildApplicationService(m *mobilesecurityservicev1alpha1.MobileSecurityService) *corev1.Service {
-	ls := getAppLabels(m.Name)
+func (r *ReconcileMobileSecurityService) buildApplicationService(service *mobilesecurityservicev1alpha1.MobileSecurityService) *corev1.Service {
+	ls := getAppLabels(service.Name)
```
laurafitzgerald: service -> mss

pkg/controller/mobilesecurityserviceapp/helpers.go
```diff
@@ -30,33 +30,33 @@ func getSdkConfigStringJsonFormat(sdk *models.SDKConfig) string {
...
+	hasFinalizer := len(mssApp.GetFinalizers()) > 0
+	isMssInstanceDeleted := service == nil
+	isMssInstanceMarkedToBeDeleted := service.GetDeletionTimestamp() != nil
 	return (isAppMarkedToBeDeleted && hasFinalizer) || isMssInstanceDeleted || isMssInstanceMarkedToBeDeleted
 }
 
 // isMobileSecurityServiceDeleted return true if it is not found because was deleted and/or was marked to be deleted
-func (r *ReconcileMobileSecurityServiceApp) isMobileSecurityServiceDeleted(operatorNamespace string, mssInstance *mobilesecurityservicev1alpha1.MobileSecurityService) bool {
-	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: utils.MobileSecurityServiceCRName, Namespace: operatorNamespace}, mssInstance); err != nil || mssInstance.GetDeletionTimestamp() != nil {
+func (r *ReconcileMobileSecurityServiceApp) isMobileSecurityServiceDeleted(operatorNamespace string, service *mobilesecurityservicev1alpha1.MobileSecurityService) bool {
```
laurafitzgerald: service -> mss , there are a couple in the following line as well.

pkg/controller/mobilesecurityservice/helpers.go
```diff
@@ -5,27 +5,29 @@ import (
...
 
 	mobilesecurityservicev1alpha1 "github.com/aerogear/mobile-security-service-operator/pkg/apis/mobilesecurityservice/v1alpha1"
-	"github.com/aerogear/mobile-security-service-operator/pkg/utils"
 	corev1 "k8s.io/api/core/v1"
 )
 
-const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+const (
+	letterBytes    = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+	oauthProxyPort = 4180
```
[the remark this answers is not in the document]
camilamacedo86: @laurafitzgerald, all that is in mandatory specs is something that can be changed in the CR. You make a previous comment that no make sense change this port, so, for this reason, it should be a local const instead of an option. I hope that it makes sense

pkg/controller/mobilesecurityservice/services.go
camilamacedo86: |o| I will look now it again.

deploy/crds/examples/mobile-security-service_v1alpha1_mobilesecurityserviceapp_cr.yaml
```diff
@@ -1,14 +1,16 @@
...
+
+# Usage: Shows how to define the object to create/update the applications in the service side
 apiVersion: mobile-security-service.aerogear.com/v1alpha1
 kind: MobileSecurityServiceApp
 metadata:
   name: mobile-security-service-app
 spec:
-  # The appName spec defines the appName of the app in the Service side, it also is used to create the ConfigMap name
-  # NOTE: To update the name just update this spec and re-apply the CR
+  # The appName should be the name of your mobile application. It will be used to create a new app in the service, the labels and the name of the ConfigMap.
```
damienomurchu:
```suggestion
  # The appName is the name of your mobile application. It is used to create a new app in the service, the labels and the name of the ConfigMap.
```

I believe the present tense (ie `is` not `will`) is preferred by the docs team, thus the suggestion (know its our repo readme and not published docs, but convention to use present tense seems to read well in docs I feel)

pkg/controller/mobilesecurityservice/helpers.go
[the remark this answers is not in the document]
laurafitzgerald: Is it not possible to have a mandatory spec then without the user specifying it? It sounds like there is no way to use the default in that case?

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
```diff
@@ -4,32 +4,52 @@ kind: MobileSecurityService
 metadata:
   name: mobile-security-service
 spec:
+
+  # NOTE: Default values will be applied by the operator if values are not specified below.
```
damienomurchu:
```suggestion
  # NOTE: Default values are applied by the operator if values are not specified below.

```

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
```diff
@@ -18,28 +18,31 @@ type MobileSecurityServiceSpec struct {
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
 
 	//Enviroment Variables for the Service
-	DatabaseName                  string `json:"databaseName"`
-	DatabasePassword              string `json:"databasePassword"`
-	DatabaseUser                  string `json:"databaseUser"`
-	DatabaseHost                  string `json:"databaseHost"`
+	DatabaseName                  string `json:"databaseName,omitempty"`
```
damienomurchu: Is it ok for this field to be empty?

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
```diff
@@ -18,28 +18,31 @@ type MobileSecurityServiceSpec struct {
...
+	DatabasePassword              string `json:"databasePassword,omitempty"`
+	DatabaseUser                  string `json:"databaseUser,omitempty"`
+	DatabaseHost                  string `json:"databaseHost,omitempty"`
```
damienomurchu: Is it ok for this field to be empty? (same as above)

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
```diff
@@ -18,28 +18,31 @@ type MobileSecurityServiceSpec struct {
...
-	Size            int32  `json:"size"`
-	Image           string `json:"image"`
-	ContainerName   string `json:"containerName"`
-	ClusterProtocol string `json:"clusterProtocol"`
-	MemoryLimit     string `json:"memoryLimit"`
-	MemoryRequest   string `json:"memoryRequest"`
-	OAuthPort       int32  `json:"oAuthPort"`
+	Size               int32  `json:"size,omitempty"`
+	Image              string `json:"image,omitempty"`
+	ContainerName      string `json:"containerName,omitempty"`
```
damienomurchu: Can all of these fields be empty?

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservicedb_types.go
```diff
@@ -20,18 +20,21 @@ type MobileSecurityServiceDBSpec struct {
...
-	Size                   int32  `json:"size"`
-	Image                  string `json:"image"`
-	ContainerName          string `json:"containerName"`
-	DatabaseMemoryLimit    string `json:"databaseMemoryLimit"`
-	DatabaseMemoryRequest  string `json:"databaseMemoryRequest"`
-	DatabaseStorageRequest string `json:"databaseStorageRequest"`
+	//CR mandatory config values
+	Size                   int32  `json:"size,omitempty"`
+	Image                  string `json:"image,omitempty"`
+	ContainerName          string `json:"containerName,omitempty"`
```
damienomurchu: OK to be empty?

pkg/controller/mobilesecurityservice/helpers.go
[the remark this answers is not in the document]
camilamacedo86: No. It is not. The mandatory_specs are to define the spec's values. If it is not a spec then has no reason for that. It never ever will be changed so the best thing is just centralized the value will a const.

pkg/controller/mobilesecurityservice/controller_test.go
```diff
@@ -296,7 +296,7 @@ func TestReconcileMobileSecurityService_Reconcile(t *testing.T) {
 	}
 
 	configMap := &corev1.ConfigMap{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: utils.GetConfigMapName(&mssInstance), Namespace: mssInstance.Namespace}, configMap)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: mssInstance.Spec.ConfigMapName, Namespace: mssInstance.Namespace}, configMap)
```
damienomurchu: Just a small question - are we better to use `utils.GetConfigMapName(&mssInstance)` so we hit the utils code & keep our test line coverage higher?

deploy/crds/examples/mobile-security-service_v1alpha1_mobilesecurityserviceapp_cr.yaml
camilamacedo86: ok.

pkg/controller/mobilesecurityservice/controller_test.go
```diff
@@ -360,7 +360,7 @@ func TestReconcileMobileSecurityService_Reconcile(t *testing.T) {
 	}
 
 	route := &routev1.Route{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: utils.GetRouteName(&mssInstance), Namespace: mssInstance.Namespace}, route)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: mssInstance.Spec.RouteName, Namespace: mssInstance.Namespace}, route)
```
damienomurchu: Same as above re utils method

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
camilamacedo86: Yes, it is since we have the mandatory_specs now.

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
camilamacedo86: Yes, it is since we have the mandatory_specs now.

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
camilamacedo86: Yes, it is since we have the mandatory_specs now.

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservicedb_types.go
camilamacedo86: Yes, it is since we have the mandatory_specs now.

pkg/controller/mobilesecurityservice/controller_test.go
```diff
@@ -431,6 +428,13 @@ func TestReconcileMobileSecurityService_Reconcile_UnknownNamespace(t *testing.T)
 	if err == nil {
 		t.Fatalf("expected not to find namespace '%v'", namespace)
 	}
+
+	// check if the deployment has been created
+	dep := &v1beta1.Deployment{}
+	err = r.client.Get(context.TODO(), req.NamespacedName, dep)
+	if err == nil {
+		t.Error("Should not create the Deployments since is an invalid namespace")
```
damienomurchu:
```suggestion
		t.Error("Should not create the Deployment since it is an invalid namespace")
```

pkg/controller/mobilesecurityservice/controller_test.go
camilamacedo86: has not need to use this func any more. It was removed from the project.

pkg/controller/mobilesecurityservice/controller_test.go
camilamacedo86: has not need to use this func any more. It was removed from the project.

pkg/controller/mobilesecurityservice/helpers.go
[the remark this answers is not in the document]
laurafitzgerald: hmm. that seems strange to me. Why have all the methods to set a default if we expect the user to supply a value each time. Maybe I'm not understanding what should be done here. Was there a JIRA/discussion somewhere with @wei-lee about what was to be done here?

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
laurafitzgerald: As we have mandatory spec, does that mean they will never be empty?

pkg/controller/mobilesecurityservice/helpers.go
```diff
@@ -37,26 +39,26 @@ func buildAppEnvVars(m *mobilesecurityservicev1alpha1.MobileSecurityService) *[]
...
+		"PGUSER":                           mss.Spec.DatabaseUser,
 	}
 }
 
 // getOAuthArgsMap is a helper to get the []string with values required/used to set OAuth for the Mobile Security Service Project
-func getOAuthArgsMap(m *mobilesecurityservicev1alpha1.MobileSecurityService) []string {
+func getOAuthArgsMap(mss *mobilesecurityservicev1alpha1.MobileSecurityService) []string {
 	return []string{
-		"--http-address=0.0.0.0:4180",
+		"--http-address=0.0.0.0:" + string(oauthProxyPort),
```
laurafitzgerald: This isn't picking it up correctly I'm getting the following when I deploy this: `--http-address=0.0.0.0:ၔ `

camilamacedo86: It was fixed there I will back as it was. More one reason for it is not in the specs :-)

pkg/controller/mobilesecurityservice/helpers.go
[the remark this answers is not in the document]
camilamacedo86: @laurafitzgerald, 

* The specs define the specs of the custom resource. ( the configuration of it )
* Something that is not configurable should not be a spec
*  The mandatory specs is just a design approach defined by @wei-lee in order to ensure that if do not specify any spec the default configuration will be applied. 

Can be the OAuthPort configurable? According to you, no. So it is not a spec.

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
damienomurchu: Saw that further down - great, will resolve this & other related comments now :)

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
damienomurchu: @laurafitzgerald will let @camilamacedo86 confirm, but it seems that way (I haven't checked all fields, but the function Camila mentioned is in this PR)

pkg/controller/mobilesecurityservicedb/test_mocks.go
```diff
@@ -32,7 +32,7 @@ var (
 		},
 	}
 
-	instanceTwo = v1alpha1.MobileSecurityServiceDB{
+	dbInstanceNotDefaultNamespace = v1alpha1.MobileSecurityServiceDB{
```
damienomurchu:
```suggestion
	dbInstanceNotDefaultNamespace = v1alpha1.MobileSecurityServiceDB{
```

pkg/controller/mobilesecurityservicedb/mandatory_specs.go
```diff
@@ -0,0 +1,97 @@
...
+	containerName          = "database"
+	databasePort           = 5432
+	databaseMemoryLimit    = "512Mi"
+	databaseMemoryRequest  = "512Mi"
+	databaseStorageRequest = "1Gi"
+)
+
+// addMandatorySpecsDefinitions will add the specs which are mandatory for Mobile Security Service CR in the case them
+// not be applied
+func addMandatorySpecsDefinitions(db *mobilesecurityservicev1alpha1.MobileSecurityServiceDB) {
```
damienomurchu: @laurafitzgerald this is the function referenced above re mandatory specs

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
camilamacedo86: The specs are defined in the CR @laurafitzgerald. If no specs are informed or not all that is mandatory for example, these are the values which will be used by default.

damienomurchu:
@camilamacedo86 PR looks good, just finished an initial review of the code - PR is very long, and for future PR's I'd suggest splitting it into multiple PRs, or keeping the commits in the PR small so they can be more easily reviewed as the PR size may slow down the speed at which it can be reviewed and approved. 

Left some small comments, but nothing that's a blocker to merging this from my initial code skim. 

As it's a large PR, I'd suggest it be verified and tested properly to ensure nothing has been inadvertently broken by any of the changes. 

I'm going to pull down the PR and verify it locally now.

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
laurafitzgerald: I think we can remove the omitifempty then as either way they will have a value in it?

damienomurchu:
@camilamacedo86 I've pulled down your changes and deployed them on my minishift environment, and can verify all functionality still works as it should. I've added the `verified` tag, but won't approve for now to give others a change to review. 

If everyone is happy with the changes, and there are no more changes to be made I'm happy to approve then.

pkg/apis/mobilesecurityservice/v1alpha1/mobilesecurityservice_types.go
camilamacedo86: NO, we cannot remove this. It is what define that the spec in CR is optional.

pkg/controller/mobilesecurityservice/helpers.go
```diff
@@ -37,26 +39,26 @@ func buildAppEnvVars(m *mobilesecurityservicev1alpha1.MobileSecurityService) *[]
...
 // getOAuthArgsMap is a helper to get the []string with values required/used to set OAuth for the Mobile Security Service Project
-func getOAuthArgsMap(m *mobilesecurityservicev1alpha1.MobileSecurityService) []string {
+func getOAuthArgsMap(mss *mobilesecurityservicev1alpha1.MobileSecurityService) []string {
 	return []string{
 		"--http-address=0.0.0.0:4180",
 		"--https-address=",
 		"--provider=openshift",
-		fmt.Sprintf("--openshift-service-account=%s", m.Name),
+		fmt.Sprintf("--openshift-service-account=%s", mss.Name),
 		"--upstream=http://localhost:3000",
```
laurafitzgerald: The port here needs to be specified if we are exposing it in the spec

camilamacedo86: It was changed.

camilamacedo86:
@laurafitzgerald really tks for your commitment and your fantastic review here and indeed your good spots.

PR #113, opened 2019-06-04
fix(image): updates operator to use quay.io mss image (AEROGEAR-9105)

damienomurchu:
## Motivation
update operator to use quay.io image for mobile-security-service

## What
update operator to use quay.io image for mobile-security-service

## Why
mobile-security-service moving to quay.io

## How
update image ref in cr file

## Verification Steps
Deploy operator, confirm it deploys correctly, and verify it is using quay.io image for mobile-security-service

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
+++ b/deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
@@ -37,7 +37,7 @@ spec:
   # Service Container
   # ---------------------------------
 
-  image: "aerogear/mobile-security-service:master"
+  image: "quay.io/aerogear/mobile-security-service:master"
   containerName: "application"
   memoryLimit: "512Mi"
   memoryRequest: "512Mi"
--- a/pkg/controller/mobilesecurityservice/mandatory_specs.go
+++ b/pkg/controller/mobilesecurityservice/mandatory_specs.go
@@ -18,7 +18,7 @@ const (
 	clusterProtocol               = "http"
 	memoryLimit                   = "512Mi"
 	memoryRequest                 = "512Mi"
-	image                         = "aerogear/mobile-security-service:master"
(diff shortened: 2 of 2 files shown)


deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
```diff
@@ -37,7 +37,7 @@ spec:
   # Service Container
   # ---------------------------------
 
-  image: "aerogear/mobile-security-service:master"
+  image: "quay.io/aerogear/mobile-security-service:master"
```
davidffrench: This should be pointing to a tag rather than master. Reduces the chance of an issue if master is somehow broken.

camilamacedo86:
The code shows fine. To test we need to set up the CR as follows. Testing it. 

 ```ymal
 # ---------------------------------
  # Service Container
  # ---------------------------------

  image: "quay.io/aerogear/mobile-security-service:master"
  containerName: "application"
  memoryLimit: "512Mi"
  memoryRequest: "512Mi"
  # Use the following spec if you would like to define the image pull policy
  containerImagePullPolicy: "Always"
```

deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
damienomurchu: @davidffrench agree, it should point to the `latest` tag, but as we don't have any releases or a latest tag yet I've left it at master for now. Once we do have a release it should be changed

camilamacedo86: Yes ..we need create a task for we do it 
to ensure that all will works

damienomurchu:
Thanks @camilamacedo86 :+1:

PR #115, opened 2019-06-11
feat(grafana-dashboard):grafana dashboard (INTLY-1458)

austincunningham:
## Motivation
https://issues.jboss.org/browse/INTLY-1895

## What
Grafana dashboard 

## Why
for Integreatly monitoring of operator 

## How
Added a yaml

## Verification Steps
 
Check the Grafana dashboard here https://grafana-route-middleware-monitoring.apps.austin-3e02.openshiftworkshop.com/d/_cxgpLnZz/mobile-security-service-operator?orgId=1&from=now%2Fd&to=now%2Fd

image

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

PS.: Add images and/or .gifs to illustrate what was changed if this pull request modifies the appearance/output of something presented to the users.

### DIFF
--- a/deploy/monitor/grafana-dashboard.yaml
+++ b/deploy/monitor/grafana-dashboard.yaml
@@ -1,0 +1,511 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  name: mobile-security-service-operator
+  labels:
+    monitoring-key: middleware
+    prometheus: application-monitoring
+spec:
+  selector:
+    matchLabels:
+      name: mobile-security-service-operator
+  json: |
+    {
+      "__inputs": [
+          {
+            "name": "DS_PROMETHEUS",
+            "label": "Prometheus",
+            "description": "",
+            "type": "datasource",
+            "pluginId": "prometheus",
+            "pluginName": "Prometheus"
+          }
+        ],
+        "__requires": [
+          {
+            "type": "grafana",
+            "id": "grafana",
+            "name": "Grafana",
+            "version": "4.3.2"
+          },
+          {
+            "type": "panel",
+            "id": "graph",
+            "name": "Graph",
+            "version": ""
+          },
+          {
+            "type": "datasource",
+            "id": "prometheus",
+            "name": "Prometheus",
+            "version": "1.0.0"
+          },
+          {
+            "type": "panel",
+            "id": "singlestat",
+            "name": "Singlestat",
+            "version": ""
+          }
+        ],
+        "annotations": {
+          "list": [
+            {
+              "builtIn": 1,
+              "datasource": "-- Grafana --",
+              "enable": true,
+              "hide": true,
+              "iconColor": "rgba(0, 211, 255, 1)",
+              "name": "Annotations & Alerts",
+              "type": "dashboard"
+            }
+          ]
+        },
+        "description": "Application metrics",
+        "editable": true,
+        "gnetId": null,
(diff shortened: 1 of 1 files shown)


davidffrench:
+1 on Camilas comment of moving to the monitoring folder.

One other change from me is the datapoint for CPU cores seems unusually with the 0.00 format.

@Nanyte25 @david-martin Would you mind taking a look over the dashboard to provide feedback please. https://grafana-route-middleware-monitoring.apps.austin-a048.openshiftworkshop.com/d/JDHB0WnZk/mobile-security-service?refresh=10s&orgId=1

david-martin:
Graphing cpu usage in when portions of a cpu could be at play e.g. if you've set limits to 500 millicores.
I haven't done much here to offer any advice yet.

For showing historical data for a boolean value (uptime) the current graph doesn't look that great.
How does it look if the system was down?

Perhaps something to consider is the Discrete plugin which makes state changes very obvious
https://grafana.com/plugins/natel-discrete-panel
and also shows the percentage up/down time for the time range being viewed.

The only suggestion for memory usage is to also show a constant line for the memory limit of the pod, if there is one set.

All just thoughts and recommendations. I would see what's there already as better than nothing

austincunningham:
@david-martin up time with outage plus some of the suggested changes. 
image

deploy/monitor/grafana-dashboard.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: integreatly.org/v1alpha1
 kind: GrafanaDashboard
 metadata:
-  name: mobile-security-service
+  name: mobile-security-service-operator
```
camilamacedo86: Should not the name of this object be related to Grafana? I suggest something as `mss-grafana-dashboard-operator`

deploy/monitor/grafana-dashboard.yaml
```diff
@@ -0,0 +1,397 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  name: mobile-security-service-operator
+  namespace: [[ .Namespace ]]
```
camilamacedo86:
```suggestion
```
It is used just in the templates. We are not using templates. We do not need to add the namespace here. It will be applied in the current namespace when the make command is executed.

deploy/monitor/grafana-dashboard.yaml
```diff
@@ -0,0 +1,397 @@
...
+                "points": false,
+                "renderer": "flot",
+                "seriesOverrides": [],
+                "spaceLength": 10,
+                "span": 12,
+                "stack": false,
+                "steppedLine": false,
+                "targets": [
+                  {
+                    "expr": "up{namespace='mobile-security-service'}",
```
camilamacedo86: Is mandatory have the namespace here? Could we do the search by the name of the pod only?
Reason: It makes harder for the user which would like to run it another namespace

deploy/monitor/grafana-dashboard.yaml
austincunningham: :+1:

deploy/monitor/grafana-dashboard.yaml
austincunningham: Thanks @camilamacedo86 , must have missed this one the rest are job. :+1:

laurafitzgerald:
Following the verification steps, the graphana dashboard is as shown in the image in the description.

PR #116, opened 2019-06-11
fix(helpers): Add name label to service

davidffrench:


pkg/controller/mobilesecurityservice/helpers.go
```diff
@@ -15,7 +15,7 @@ const (
 
 //getAppLabels returns an string map with the labels which wil be associated to the kubernetes/ocp resource which will be created and managed by this operator
 func getAppLabels(name string) map[string]string {
-	return map[string]string{"app": "mobilesecurityservice", "mobilesecurityservice_cr": name}
+	return map[string]string{"app": "mobilesecurityservice", "mobilesecurityservice_cr": name, "name": "mobilesecurityservice"}
```
camilamacedo86:
```suggestion
	return map[string]string{"app": "mobilesecurityservice", "mobilesecurityservice_cr": name, "name": name}
```
Could you not use the label "app"?

PR #124, opened 2019-06-14
fix(script) : Remove AEROGEAR check from the commit msg

camilamacedo86:
## Motivation
Unable to use the commit hooks when the task is not tracked in the AEROGEAR

## What
- Remove the check that makes be mandatory have the "AEROGEAR-*****" from the checks

## Why
For example, now we are doing tasks tracked in the integr8ly project for it. In this way, no make sense we have this check and in my pov, it has not been so helpful at all. 

## How
- just remove the specify checks for it from the script called by the git hook

## Verification Steps
Do any git operation which will call it and do not add AEROGEAR in the msg
Following the test performed. 
 
Screenshot 2019-06-14 at 10 30 24

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/scripts/commit-filter-check.sh
+++ b/scripts/commit-filter-check.sh
@@ -14,41 +14,31 @@ commit_message_check (){
       gitmessage=`git log --format=%B -n 1 "$i"`
       
       ####################### TEST STRINGS comment out line 13 to use #########################################
-      #gitmessage="feat sdasdsadsaas (AEROGEAR-asdsada)"
-      #gitmessage="feat(some txt): some txt (AEROGEAR-****)"
-      #gitmessage="docs(some txt): some txt (AEROGEAR-1234)"
-      #gitmessage="fix(some txt): some txt (AEROGEAR-5678)"
+      #gitmessage="feat sdasdsadsaas "
+      #gitmessage="feat(some txt): some txt "
+      #gitmessage="docs(some txt): some txt"
+      #gitmessage="fix(some txt): some txt"
       #########################################################################################################
-      
+
       messagecheck=`echo $gitmessage | grep -w "feat\|fix\|docs\|breaking"`
       if [ -z "$messagecheck" ]
-      then 
+      then
             echo "Your commit message must begin with one of the following"
             echo "  feat(feature-name)"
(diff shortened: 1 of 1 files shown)


scripts/commit-filter-check.sh
```diff
@@ -29,26 +29,6 @@ commit_message_check (){
...
       fi
-      messagecheck=`echo $gitmessage | grep "(AEROGEAR-"`
-      if  [ -z "$messagecheck" ]
-      then 
-            echo "Your commit message must end with the following"
-            echo "  (AEROGEAR-****)"
-            echo "Where **** is the Jira number"
-            echo " " 
-      fi
-      messagecheck=`echo $gitmessage | grep ": "`
```
davidffrench: This is removing the need for `:` as well. Why do we want to do this?

camilamacedo86: I will check

camilamacedo86: Done .

damienomurchu:
+1 for removing it if there is a potential scenario where a PR might not have an associated JIRA (contributions from the community?).

Agree with @davidffrench though, it is useful to have the ticket reference in there when it is available

camilamacedo86:
@davidffrench The point is that we cannot force the number of the ticket and the format. It indeed could be an ISSUE from the repo. Is it make sense? If not, could you describe how we can move forward here? 

c/c @damienomurchu

davidffrench:
I would like to keep the ticket reference in the commit. It is unlikely to include another JIRA project and even if it does, we can add that. So I am happy to have it changed in the same way as https://github.com/aerogear/mobile-security-service/pull/161/files

camilamacedo86:
Hi @davidffrench,

But this solution shows not good at all since it requires a change in the script for each new project which can track tasks that will be required to be made here and mainly because it also not allow we add the ISSUES tracked in the repo.

davidffrench:
@camilamacedo86 I am happy for it to be removed also. There will likely be limited changes the repo moving forward.

camilamacedo86:
Hi, @davidffrench it sounds great !!! Could you approve this PR so?

PR #127, opened 2019-06-17
fix(log) : Add validation for when the error faced when the service is called did not result in an HTTP response 

camilamacedo86:
## Motivation
The scenario was checked by running the app locally. 

## What
Add validation for when the error faced when the service is called did not result in an HTTP response 

## Why
Avoid panic because the response is nil for example when debugging it locally. 

## Verification Steps
Check the CI

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/pkg/service/client.go
+++ b/pkg/service/client.go
@@ -27,8 +27,11 @@ var DeleteAppFromServiceByRestAPI = func(serviceAPI string, id string, reqLogger
 	client := &http.Client{}
 	response, err := client.Do(req)
 
-	if err != nil || 204 != response.StatusCode {
-		reqLogger.Error(err, "HTTP StatusCode not expected", "HTTPMethod", http.MethodDelete, "url", url, "response.StatusCode", response.StatusCode)
+	if err != nil || response == nil || 204 != response.StatusCode {
+		if response != nil {
+			reqLogger.Error(err, "Unable to execute GET request", "HTTPMethod", http.MethodGet, "url", url, "response.StatusCode", response.StatusCode)
+		}
+		reqLogger.Error(err, "Unable to execute GET request", "HTTPMethod", http.MethodGet, "url", url)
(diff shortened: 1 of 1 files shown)


pkg/service/client.go
```diff
@@ -28,7 +28,10 @@ var DeleteAppFromServiceByRestAPI = func(serviceAPI string, id string, reqLogger
 	response, err := client.Do(req)
 
 	if err != nil || 204 != response.StatusCode {
```
davidffrench: @camilamacedo86 If response is nil, this line of `response.StatusCode` will just throw a runtime error.

camilamacedo86: Yes. I will fix it properly and ping you.

camilamacedo86:
@laurafitzgerald @austincunningham could you please check this one for we move forward here.

pkg/service/client.go
```diff
@@ -99,7 +105,10 @@ func GetAppFromServiceByRestApi(serviceAPI string, appId string, reqLogger logr.
 	client := &http.Client{}
 	response, err := client.Do(req)
 	if err != nil {
-		reqLogger.Error(err, "Unable to execute GET request", "HTTPMethod", http.MethodGet, "url", url, "response.StatusCode", response.StatusCode)
+		if response != nil {
+			reqLogger.Error(err, "Unable to execute GET request", "HTTPMethod", http.MethodGet, "url", url, "response.StatusCode", response.StatusCode)
+		}
+		reqLogger.Error(err, "Unable to execute GET request", "HTTPMethod", http.MethodGet, "url", url)
```
austincunningham: not sure you need to do this here as nil is being returned

camilamacedo86: When the issue occurs the response is nil. In this case, the response can be nil which will cause panic.

PR #133, opened 2019-06-24
Small Fix  - fix(db) : status

camilamacedo86:
## Motivation
Issue found 

## What
The ServiceStatus was not been updated. 

## Verification Steps
1. Check the CI
2. You can also use the image `docker.io/cmacedo/mobile-security-service-operator:serviceupdatefix
` to check that the status should be added in the Resource in the OCP. 

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/pkg/controller/mobilesecurityservicedb/status.go
+++ b/pkg/controller/mobilesecurityservicedb/status.go
@@ -110,6 +110,7 @@ func (r *ReconcileMobileSecurityServiceDB) updateServiceStatus(reqLogger logr.Lo
 
 		// Set the data
 		instance.Status.ServiceName = serviceStatus.Name
+		instance.Status.ServiceStatus = serviceStatus.Status
 
 		// Update the CR
 		err = r.client.Status().Update(context.TODO(), instance)


laurafitzgerald:
@camilamacedo86 I have the changes deployed, can you specify exactly where you would expect the `the status should be added in the Resource in the OCP.` I've checked the db deployment and it has a status:
```
status:
  availableReplicas: 1
  fullyLabeledReplicas: 1
  observedGeneration: 3
  readyReplicas: 1
  replicas: 1
```

laurafitzgerald:
@camilamacedo86 I was also able to see that the status of the cr changed if I tried to scale down the db pod. I have posted a video to the chat channel to show this.

camilamacedo86:
Hi @laurafitzgerald,

As we spoke in the video sent for you we can check the Service Status there which is the fix here. It was not been set. If you try scale the db the operator will be trigge1 and will set the size according to the spec again .. the status will change because the status are dynamic. However, when all be finished and ok again the changes will stop.

camilamacedo86:
really tks @laurafitzgerald

PR #134, opened 2019-06-25
fix(url): removed `/init` from URL

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-9486 

## What

Removed `/init` from the URL that is added to the SDKConfig

## Why

This is added by the JS-SDK. 

## Verification Steps
 
1. Replace the value for `template.spec.containers.image` with `quay.io/ephelan/mobile-security-service-operator:AEROGEAR-9486`
2. Run `make install`.
3. Run `make example-app/apply`.
4. Run `oc get cm example-appname-security -o json`. The value at `data.SDKConfig.url` should be the root URL, without an `/init` amended to it. Like this:

```sh
❯ oc get cm example-appname-security -o json                                                                                                                                 
{                                                                                                                                                                            
    "apiVersion": "v1",                                                                                                                                                      
    "data": {                                                                                                                                                                
        "SDKConfig": "{\n\t\"url\": \"http://route-mobile-security-service.192.168.42.115.nip.io\"\n}"                                                                       
    },
	...                                                                                                                                                                                                                                                                                                                                             
}
```

## Checklist:

- [x] Changes have been successfully verified by another team member 

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/pkg/controller/mobilesecurityserviceapp/controller.go
+++ b/pkg/controller/mobilesecurityserviceapp/controller.go
@@ -210,13 +210,13 @@ func (r *ReconcileMobileSecurityServiceApp) Reconcile(request reconcile.Request)
 		return reconcile.Result{}, err
 	}
 
-	// Get the Public Service API URL which will be used to build the SDKConfigMap json
-	publicServiceURLAPI := utils.GetInitPublicURL(route, mss)
+	// Get the Public URL which will be used to build the SDKConfigMap json
+	apiURL := utils.GetPublicURL(route, mss)
 
 	reqLogger.Info("Checking if the configMap already exists ...")
 	// Check if ConfigMap for the app exist, if not create one.
 	if _, err := r.fetchConfigMap(reqLogger, mssApp); err != nil {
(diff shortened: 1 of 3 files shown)


camilamacedo86:
Why you are removing it @craicoverflow? It was requested before.  The SDK will not use any other endpoint and it is as defined.

craicoverflow:
> Hi @craicoverflow,
> 
> I am ok with removing it. However, why you are removing it? My concern is that it was requested before. The SDK will not use any other endpoint and it is as defined. Are you sure over it? If yes, go forward.

There are two reasons why it should be the root URL and not the `/init` endpoint. Myself, @laurafitzgerald and @austincunningham realised this while you were away.

1. The MDC requires the root URL as it will serve as a link that the user can click on to open the Mobile Security Service UI.
2. **aerogear-js-sdk** actually appends `/api/init` to the URL for the app. See [AppSecurity.ts#L53](https://github.com/aerogear/aerogear-js-sdk/blob/c279617f7d0f4f2fdd387f6010bfe169bb461289/packages/security/src/appSecurity/AppSecurity.ts#L53).

PR #136, opened 2019-06-26
fix(Makefile) : Improve output and fix order for uninstall

camilamacedo86:
## What
- improve outputs of make install/uninstall
- move `kubectl delete -f deploy/operator.yaml` for after the CRD's (it cannot be done before)


Makefile
```diff
@@ -30,30 +30,36 @@ LDFLAGS=-ldflags "-w -s -X main.Version=${TAG}"
...
+	@echo ....... Deleting the Mobile Security Service and Database .......
 	- kubectl delete -f deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml
 	- kubectl delete -f deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservicedb_cr.yaml
-	- kubectl delete -f deploy/operator.yaml
-	@echo ....... Delete Operator and Service....... 
+	@echo ....... Deleting CRDs.......
 	- kubectl delete -f deploy/crds/mobile-security-service_v1alpha1_mobilesecurityserviceapp_crd.yaml
 	- kubectl delete -f deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_crd.yaml
 	- kubectl delete -f deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservicedb_crd.yaml
+	@echo ....... Deleting Operator.......
```
laurafitzgerald: Hi @camilamacedo86 can I ask about this change. How does the removal of the crds after the removal of the operator cause a problem?

camilamacedo86: because of the finalizer.

laurafitzgerald: Thanks @camilamacedo86 Do the steps here need to be verified?

camilamacedo86: Just is still working the make file. It is very small, IMHO we should not spend too much time on it.

PR #138, opened 2019-06-27
adding operator sops

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/INTLY-1894

## What
Adding SOPs from doc

### DIFF
--- a/SOP/SOP-mss.md
+++ b/SOP/SOP-mss.md
@@ -1,1 +1,108 @@
 # Mobile Security Service - Standard Operating Procedures
+
+# Critical
+
+## MobileSecurityServiceConsoleDown
+
+Troubleshoot using the following steps via either the console or the cli:
+
+Console:
+1. Check the operator pod is present as it is responsible for managing the service pod as described in [MobileSecurityServiceOperatorDown](https://github.com/aerogear/mobile-security-service-operator/SOP/SOP-operator.md)
+    1. If resolving the MobileSecurityServiceOperatorDown doesn't resolve the issue, please continue with the below steps
+
+2. Check that the Service Pod is deployed in the same namespace as the operator
+3. Check the status of the Mobile Security Service Custom Resources
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service -> mobile-security-service -> Actions -> Edit yaml
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+4. Check the status of the Mobile Security Service DB Custom Resource. If the Database Pod is not available this will cause the Service pod to error
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service DB-> mobile-security-service-db -> Actions -> Edit yaml
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+6. If the service pod is present check the logs of the OAuth Proxy Container
+    1. Go To -> Applications -> Pods -> mobile-security-service-<xyz123> -> Logs -> Container -> oauth-proxy
+7. If the service pod is present check the logs of the Application Container
+    1. Go To -> Applications -> Pods -> mobile-security-service-<xyz123> -> Logs -> Container -> application
+
+
+CLI:
+1. Check the operator pod is present as it is responsible for managing the service pod as described in [MobileSecurityServiceOperatorDown](https://github.com/aerogear/mobile-security-service-operator/SOP/SOP-operator.md)
+    1. If resolving the MobileSecurityServiceOperatorDown doesn't resolve the issue, please continue with the below steps
+
+2. Check that the Service Pod is deployed in the same namespace as the operator
+2. Check the status of the Mobile Security Service Custom Resources
+    1. `oc get MobileSecurityService mobile-security-service -o yaml`
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+3. Check the status of the Mobile Security Service DB Custom Resource. If the Database Pod is not available this will cause the Service pod to error
+    1. `oc get MobileSecurityServiceDB mobile-security-service-db -o yaml`
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+6. If the service pod is present check the logs of the OAuth Proxy Container
+    1. Get the service pod name -> `oc get pods | grep mobile-security-service`
+    2. `oc logs <service-podname> -c oauth-proxy`
+    3. Save the logs by running `oc logs <service-podname> -c oauth-proxy > <filename>.log`
+7. If the service pod is present check the logs of the Application Container
+    1. `oc logs <service-podname> -c application`
+    2. Save the logs by running `oc logs <service-podname> -c application > <filename>.log`
+
+
+## MobileSecurityServiceDown
+
+1. Check the steps for the [MobileSecurityServiceConsoleDown](https://github.com/aerogear/mobile-security-service-operator/SOP/SOP-operator.md) in order to troubleshoot this alert
+
+## MobileSecurityServiceDatabaseDown
+
+Troubleshoot using the following steps via either the console or the cli:
+
+Console:
+1. Check the operator pod is present as it is responsible for managing the service pod as described in [MobileSecurityServiceOperatorDown](https://github.com/aerogear/mobile-security-service-operator/SOP/SOP-operator.md)
+    1. If resolving the MobileSecurityServiceOperatorDown doesn't resolve the issue, please continue with the below steps
+
+2. Check that the Database Pod is deployed in the same namespace as the operator
+3. Check the status of the Mobile Security Service DB Custom Resource. If the Database Pod is not available this will cause the Service pod to error
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service DB-> mobile-security-service-db -> Actions -> Edit yaml
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+4. Check that the DB pod is using values correct from a config map
+    1. Navigate to Resources -> Config Maps -> mobile-security-service-config to ensure this config map exists and contains values for ACCESS_CONTROL_ALLOW_CREDENTIALS, ACCESS_CONTROL_ALLOW_ORIGIN, LOG_FORMAT, LOG_LEVEL, PGDATABASE, PGHOST, PGPASSWORD, PGUSER.
+    2. Navigate to Applications -> Deployments -> mobile-security-service-db -> Environment
+    3. Verify that the Container dtabase is sourcing POSTGRESQL_DATABASE, POSTGRESQL_USER, POSTGRESQL_PASSWORD from the config map mobile-security-service-config Config Map
+3. If the DB pod is present check the logs of the pod for any errors
+    1. Navigate to Applications -> Pods -> mobile-security-service-db-<xyz123> -> Logs
+
+
+CLI:
+1. Check the operator pod is present as it is responsible for managing the service pod as described in [MobileSecurityServiceOperatorDown](https://github.com/aerogear/mobile-security-service-operator/SOP/SOP-operator.md)
+    1. If resolving the MobileSecurityServiceOperatorDown doesn't resolve the issue, please continue with the below steps
+
+2. Check that the Database Pod is deployed in the same namespace as the operator
+2. Check the status of the Mobile Security Service DB Custom Resource. If the Database Pod is not available this will cause the Service pod to error
+    1. `oc get MobileSecurityServiceDB mobile-security-service-db -o yaml`
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+3. If the DB pod is present check the logs of the pod for any errors
+    1. Get the service pod name -> `oc get pods | grep mobile-security-service-db`
+    2. `oc logs <database-podname>`
+    3. Save the logs by running `oc logs <database-podname> > <filename>.log`
+
+# Warning
+
+## MobileSecurityServicePodCPUHigh ## MobileSecurityServicePodMemoryHigh
+
+1. Capture a snapshot of the 'Mobile Security Service Application' Grafana dashboard and track it over time. The metrics can be useful for identifying performance isssues over time.
+
+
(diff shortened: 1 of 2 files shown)


SOP/SOP-operator.md
```diff
@@ -1 +1,69 @@
...
+2. Switch to the Mobile Security Service namespace
+3. Check the status of the Mobile Security Service Custom Resources
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service -> mobile-security-service -> Actions -> Edit yaml -> Scroll Down to status -> appStatus 
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+4. Check the status of the Mobile Security Service DB Custom Resource
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service DB-> mobile-security-service-db -> Actions -> Edit yaml -> Scroll Down to status -> appStatus 
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+5. If the operator pod is present check the logs of the pod for any errors
+    1. Navigate to Applications -> Pods -> mobile-security-service-operator-<xyz123> -> Logs
+    2. Check for any errors in the logs
```
austincunningham: maybe add check the events here

austincunningham: Never mind see it below

SOP/SOP-operator.md
```diff
@@ -1 +1,69 @@
...
+## MobileSecurityServiceOperatorDown
+
+Troubleshoot using the following steps via either the console or the cli:
+
+Console:
+1. Log into the OpenShift console
+2. Switch to the Mobile Security Service namespace
+3. Check the status of the Mobile Security Service Custom Resources
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service -> mobile-security-service -> Actions -> Edit yaml -> Scroll Down to status -> appStatus 
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
```
camilamacedo86: Wdty over we highlight that all that is created and required for MSS has a status in the table? Maybe a NOTE. It is just an idea

SOP/SOP-operator.md
```diff
@@ -1 +1,69 @@
...
+4. Check the status of the Mobile Security Service DB Custom Resource
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service DB-> mobile-security-service-db -> Actions -> Edit yaml -> Scroll Down to status -> appStatus
```
camilamacedo86: The name of the status is not accurate.

SOP/SOP-operator.md
```diff
@@ -1 +1,69 @@
...
+
+Console:
+1. Log into the OpenShift cluster
+2. Switch to the Mobile Security Service namespace
+3. Navigate to the Overview using the side navigation
+4. Check the following
+    1. There is one pod running under the application mobilesecurityservice
+    2. There is one pod running under the application mobilesecurityservicedb
+    3. There is one pod running under the Other Resources
+5. The operator should  maintain the number of pods running for the mobilesecurity service, please resolve the operator pod numbers first by scaling it up or down as necessary.
```
austincunningham:
```suggestion
5. The operator should  maintain the number of pods running for the mobile security service, please resolve the operator pod numbers first by scaling it up or down as necessary.
```

SOP/SOP-operator.md
```diff
@@ -1 +1,69 @@
...
+1. Log into the OpenShift console
+2. Switch to the Mobile Security Service namespace
+3. Check the status of the Mobile Security Service Custom Resources
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service -> mobile-security-service -> Actions -> Edit yaml -> Scroll Down to status -> appStatus 
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+4. Check the status of the Mobile Security Service DB Custom Resource
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service DB-> mobile-security-service-db -> Actions -> Edit yaml -> Scroll Down to status -> appStatus 
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+5. If the operator pod is present check the logs of the pod for any errors
+    1. Navigate to Applications -> Pods -> mobile-security-service-operator-<xyz123> -> Logs
```
camilamacedo86: - We can add here to check if the Env Var values are mapped to use the ConfigMap created by MSS CR
- We should recommend checking here the logs of the database pod before check the operator one.

SOP/SOP-operator.md
```diff
@@ -1 +1,69 @@
...
+    2. Check for any errors in the logs
+6. If the operator pod is not present scale up the pod
+    1. Navigate to Overview -> Click Dropdown for Deployment mobile-security-service-operator -> Click on up arrow to scale the pod to 1
```
camilamacedo86: It is a scenario that will never occur. The Operator ensures that the qt of sizes spec will be scaled. See here https://github.com/aerogear/mobile-security-service-operator/blob/master/pkg/controller/mobilesecurityservicedb/controller.go#L184

If the admin/user try to scale it for up or down the operator will fix automatically and it will take less than 5 min.

SOP/SOP-operator.md
```diff
@@ -1 +1,69 @@
...
+Console:
+1. Log into the OpenShift console
+2. Switch to the Mobile Security Service namespace
+3. Check the status of the Mobile Security Service Custom Resources
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service -> mobile-security-service -> Actions -> Edit yaml -> Scroll Down to status -> appStatus 
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+4. Check the status of the Mobile Security Service DB Custom Resource
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service DB-> mobile-security-service-db -> Actions -> Edit yaml -> Scroll Down to status -> appStatus 
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+5. If the operator pod is present check the logs of the pod for any errors
```
camilamacedo86: If the operator pod is not present then is not possible to scale it and it means that the operator.yaml was deleted.
IMHO: If any POD be not present, I mean do not exist, the step is `make install` to apply the CR/CRD which will create each one. In the case of operator pod, it is the operator.yaml file which needs be reapplied.

SOP/SOP-operator.md
camilamacedo86: - Check if the DB is created and running OK
- Check the logs of MSS service which is running in this pod. ( maybe it is not able to connect in the database )
- Check in the ENV var of the DB if it is using the ConfigMap values created by MSS ( maybe the databasename, user, and pwd used are not that ones which are expected because someone change it manually )

SOP/SOP-operator.md
```diff
@@ -1 +1,69 @@
...
+
+Console:
+1. Log into the OpenShift console
+2. Switch to the Mobile Security Service namespace
+3. Check the status of the Mobile Security Service Custom Resources
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service -> mobile-security-service -> Actions -> Edit yaml -> Scroll Down to status -> appStatus 
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+4. Check the status of the Mobile Security Service DB Custom Resource
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service DB-> mobile-security-service-db -> Actions -> Edit yaml -> Scroll Down to status -> appStatus 
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
```
camilamacedo86: - Check values used in the ENV Var 
- Check db pod logs

SOP/SOP-operator.md
```diff
@@ -1 +1,69 @@
 # Mobile Security Service Operator- Standard Operating Procedures
+
+## MobileSecurityServiceOperatorDown
+
+Troubleshoot using the following steps via either the console or the cli:
+
+Console:
+1. Log into the OpenShift console
```
camilamacedo86: For MSS and DB check if are deployed in the same namespace of the Operator.

SOP/SOP-operator.md
laurafitzgerald: It is possible to scale down the operator pod manually. So there is a case where it could be not there.

SOP/SOP-operator.md
laurafitzgerald: This SOP is for the operator being down. I will add these ones to another check.

SOP/SOP-operator.md
laurafitzgerald: Again I don't think it would be necessary to check all of these items for the operator pod down but I will include them elsewhere.

SOP/SOP-operator.md
laurafitzgerald: Yes, well spotted, I will remove it as the list which is linked shows which status' need to be checked.

SOP/SOP-operator.md
laurafitzgerald: I've included steps below to scale the pod. It is possible for there not to be an operator pod and for the operator.yaml file to still be there.

SOP/SOP-mss.md
```diff
@@ -1 +1,99 @@
...
+    1. If resolving the MobileSecurityServiceOperatorDown doesn't resolve the issue, please continue with the below steps
+
+2. Check that the Database Pod is deployed in the same namespace as the operator
+2. Check the status of the Mobile Security Service DB Custom Resource. If the Database Pod is not available this will cause the Service pod to error
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service DB-> mobile-security-service-db -> Actions -> Edit yaml
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+3. If the DB pod is present check the logs of the pod for any errors
+    1. Navigate to Applications -> Pods -> mobile-security-service-db-<xyz123> -> Logs
+    2. Check for any errors in the logs
+
```
camilamacedo86: It is missing just check the ENV VAR of the DB pod to see if it is using the ConfigMap and/or if the values used are correct.

SOP/SOP-mss.md
```diff
@@ -1 +1,99 @@
...
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+3. If the DB pod is present check the logs of the pod for any errors
+    1. Get the service pod name -> `oc get pods | grep mobile-security-service-db`
+    2. `oc logs <database-podname>`
+    3. Save the logs by running `oc logs <database-podname> > <filename>.log`
+
+# Warning
+
+## MobileSecurityServicePodCPUHigh
+
```
camilamacedo86: Could increase the instances. However, for do that is required to change the MSS CR spec size 1 to X. Will we allow this action?

SOP/SOP-mss.md
```diff
@@ -1 +1,99 @@
...
+
+
+
+
+## MobileSecurityServicePodMemoryHigh
+
+
```
camilamacedo86: Could be solved by increase the instances(scale the pod). However, for do that is required to change the MSS CR spec size 1 to X. Will we allow this action?

SOP/SOP-mss.md
```diff
@@ -1 +1,99 @@
...
+
+
+## MobileSecurityServiceApiHighRequestDuration
+
```
camilamacedo86: I cannot see a solution for the admin. Maybe here is change the log level of the Service via the ENV VAR of its pod to debug, get the data and contact us.

SOP/SOP-mss.md
```diff
@@ -1 +1,99 @@
...
+
+
+## MobileSecurityServiceApiHighConcurrentRequests
+
+## MobileSecurityServiceApiHighRequestFailure
```
camilamacedo86: Maybe here is change the log level of the Service via the ENV VAR of its pod to debug and check

SOP/SOP-mss.md
laurafitzgerald: Yes I think that's a good suggestion

laurafitzgerald: I'm not aware of a way that we can do this. There is no way to trigger new rollout of a deployment manually. Changing the value in the config map alone doesn't trigger a rollout. It would be possible if we use deployment configs instead of deployments.

SOP/SOP-mss.md
```diff
@@ -1 +1,108 @@
...
+
+
+## MobileSecurityServiceApiHighRequestDuration, MobileSecurityServiceApiHighRequestFailure , MobileSecurityServiceApiHighConcurrentRequests
+
+1. Capture a snapshot of the Grafana dashboard. The metrics can be useful for identifying errors and performance bottlenecks.
+
+Troubleshoot using the following steps via either the console or the cli:
+
+Console:
+1. Capture a snapshot of the 'Fuse Online - Infrastructure - REST APIs' Grafana dashboard. The metrics can be useful for identifying errors and performance bottlenecks.
```
austincunningham: Don't think we need to reference Fuse here

SOP/SOP-mss.md
```diff
@@ -1 +1,108 @@
...
+    1. If resolving the MobileSecurityServiceOperatorDown doesn't resolve the issue, please continue with the below steps
+
+2. Check that the Service Pod is deployed in the same namespace as the operator
+3. Check the status of the Mobile Security Service Custom Resources
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service -> mobile-security-service -> Actions -> Edit yaml
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+4. Check the status of the Mobile Security Service DB Custom Resource. If the Database Pod is not available this will cause the Service pod to error
+    1. Go To -> Resources -> Other Resources -> Choose a resource to list -> Mobile Security Service DB-> mobile-security-service-db -> Actions -> Edit yaml
+    2. Check that the status' are as expected as described in the [README](https://github.com/aerogear/mobile-security-service-operator#status-definition-per-types)
+6. If the service pod is present check the logs of the OAuth Proxy Container
```
camilamacedo86: If the DB is up and working but the service is not:
- Check the Env Var values used in the Service
- Check if all values are from the configMap if not fixed it / or delete and re-apply the MSS CR it should solve either
- Check if the DB is using the values from the shared confgMap.

PR #151, opened 2019-07-12
Fix prometheus rules and improve/fix the SOP

camilamacedo86:
## Motivation
- https://issues.jboss.org/browse/INTLY-2709 ( raised after discussions over UPS SOP and Prometheus rules and the review over it)

## What
- Remove the CLI steps from the SOP's
- Small fixes in the steps
- Fix the MobileSecurityServicePodCount ( it should alert when the pods are < 3 and NOT != of 3 )
- Fix the MobileSecurityServiceDown ( it was not working at all and shown the alert when the MSS pod was up and running )
- Add steps to scale the MSS pods in the case of some issues regards performance be faced.
- Add further information in order to attend the OPS suggestions/requirements

## Verification Steps
- Check the Sops [MMS](https://github.com/aerogear/mobile-security-service-operator/blob/3729f93bb71b836efd8ed6a25d1869231ef5339f/SOP/SOP-mss.adoc) and [Oper](https://github.com/aerogear/mobile-security-service-operator/blob/3729f93bb71b836efd8ed6a25d1869231ef5339f/SOP/SOP-operator.adoc)

- Check here the Prometheus Rule changed: https://prometheus-route-middleware-monitoring.apps.london-4bc4.openshiftworkshop.com/alerts

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes
Screenshot 2019-07-12 at 08 12 13

### DIFF
--- a/deploy/monitor/mss_prometheus_rule.yaml
+++ b/deploy/monitor/mss_prometheus_rule.yaml
@@ -15,7 +15,7 @@ spec:
     - name: general.rules
       rules:
       - alert: MobileSecurityServiceDown
-        expr: absent(up{job="mobile-security-service-application"} == 1)
+        expr: absent(kube_pod_container_status_running{namespace="mobile-security-service",container="application"}==1)
         for: 5m
         labels:
           severity: critical
--- a/deploy/monitor/mss_prometheus_rule.yaml
+++ b/deploy/monitor/mss_prometheus_rule.yaml
@@ -15,7 +15,7 @@ spec:
     - name: general.rules
       rules:
       - alert: MobileSecurityServiceDown
-        expr: absent(kube_pod_container_status_running{namespace="mobile-security-service",container="application"}==1)
+        expr: absent(kube_pod_container_status_running{namespace="mobile-security-service",container="application"}>=1)
         for: 5m
         labels:
           severity: critical
@@ -24,7 +24,7 @@ spec:
           summary: "The mobile-security-service is down. For more information see on the MSS at https://github.com/aerogear/mobile-security-service"
           sop_url: "https://github.com/aerogear/mobile-security-service-operator/blob/0.2.0/SOP/SOP-mss.adoc"
       - alert: MobileSecurityServiceConsoleDown
-        expr: absent(kube_endpoint_address_available{endpoint="mobile-security-service-application"} == 1)
+        expr: absent(kube_endpoint_address_available{endpoint="mobile-security-service-application"} >= 1)
         for: 5m
         labels:
           severity: critical
--- a/deploy/monitor/prometheus_rule.yaml
+++ b/deploy/monitor/prometheus_rule.yaml
@@ -25,8 +25,8 @@ spec:
           sop_url: "https://github.com/aerogear/mobile-security-service-operator/blob/0.2.0/SOP/SOP-operator.adoc"
       - alert: MobileSecurityServicePodCount
         annotations:
-          description: The Pod count for the mobile-security-service has changed in the last 5 minutes and has not the minimal quantity required.
-          summary: Pod count for namespace mobile-security-service is {{ printf "%.0f" $value }}. Expected 3 pods. For more information see on the MSS operator https://github.com/aerogear/mobile-security-service-operator"
+          description: "The Pod count for the mobile-security-service has changed in the last 5 minutes and is less than the minimal required value."
+          summary: "Pod count for namespace mobile-security-service is {{ printf '%.0f' $value }}. Expected 3 pods. For more information see on the MSS operator https://github.com/aerogear/mobile-security-service-operator"
           sop_url: "https://github.com/aerogear/mobile-security-service-operator/blob/0.2.0/SOP/SOP-operator.adoc"
         expr: |
           (1-absent(kube_pod_status_ready{condition="true", namespace="mobile-security-service"})) or sum(kube_pod_status_ready{condition="true", namespace="mobile-security-service"}) < 3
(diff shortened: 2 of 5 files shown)


deploy/monitor/mss_prometheus_rule.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     - name: general.rules
       rules:
       - alert: MobileSecurityServiceDown
-        expr: absent(up{job="mobile-security-service-application"} == 1)
+        expr: absent(kube_pod_container_status_running{namespace="mobile-security-service",container="application"}==1)
         for: 5m
```
camilamacedo86: It was wrong and not working at all. 
c/c @wei-lee @grdryn

deploy/monitor/prometheus_rule.yaml
```diff
@@ -23,13 +23,13 @@ spec:
...
-      - alert: MobileSecurityServicePodcount
+      - alert: MobileSecurityServicePodCount
         annotations:
-          description: The Pod count for the mobile-security-service has changed in the last 5 minutes.
+          description: The Pod count for the mobile-security-service has changed in the last 5 minutes and has not the minimal quantity required.
           summary: Pod count for namespace mobile-security-service is {{ printf "%.0f" $value }}. Expected 3 pods. For more information see on the MSS operator https://github.com/aerogear/mobile-security-service-operator"
           sop_url: "https://github.com/aerogear/mobile-security-service-operator/blob/0.2.0/SOP/SOP-operator.adoc"
         expr: |
-          (1-absent(kube_pod_status_ready{condition="true", namespace="mobile-security-service"})) or sum(kube_pod_status_ready{condition="true", namespace="mobile-security-service"}) != 3
+          (1-absent(kube_pod_status_ready{condition="true", namespace="mobile-security-service"})) or sum(kube_pod_status_ready{condition="true", namespace="mobile-security-service"}) < 3
```
wei-lee: rather than hard coding the namespace value here, is it possible to get the namespace name dynamically? The problem is that the namespace name could be different on different clusters.

cc @grdryn @laurafitzgerald are we using hardcoded namespace in UPS too?

camilamacedo86: I understand that it is done in the integr8ly scripts as it was done here for example : https://github.com/integr8ly/installation/pull/687/files#diff-eaa212d002e12d63341d8140d3803235R18.

grdryn: Ideally for the monitoring resources for the service (rather than the operator itself (although maybe it could provision for itself also, like it provisions its own Service for metrics exposure), the operator would be provisioning those (after checking that the relevant CRDs exist on the cluster already). I think that's the best eventual solution (maybe not necessary to do for this release, and if you're correct about the integreatly installer, then that's probably enough for now).

camilamacedo86: The keylock is not managed and doing the things in the way that you are proposing here as well. However, in the future maybe it could be improved horizontally in all projects as you suggested. For now, IMHO shows fine move within a way that will work for the current repo and will allow us to use it to create easily the templates to manage it by ansible scripts as it shows ben doing so far.

camilamacedo86:
Hi @wei-lee and @grdryn after we chat internally I understand that we will follow this approach for now as we agreed so please let me know if we are able to merge it now.

deploy/monitor/prometheus_rule.yaml
camilamacedo86: I understand that we agreed to move forward with this approach so far.

deploy/monitor/mss_prometheus_rule.yaml
wei-lee: should this be `>= 1` ?

deploy/monitor/prometheus_rule.yaml
```diff
@@ -23,13 +23,13 @@ spec:
           description: "The mobile-security-service-operator has been down for more than 5 minutes. "
           summary: "The mobile-security-service-operator is down. For more information see on the MSS operator https://github.com/aerogear/mobile-security-service-operator"
           sop_url: "https://github.com/aerogear/mobile-security-service-operator/blob/0.2.0/SOP/SOP-operator.adoc"
-      - alert: MobileSecurityServicePodcount
+      - alert: MobileSecurityServicePodCount
         annotations:
-          description: The Pod count for the mobile-security-service has changed in the last 5 minutes.
+          description: The Pod count for the mobile-security-service has changed in the last 5 minutes and has not the minimal quantity required.
```
wei-lee: The Pod count for the mobile-security-service has changed in the last 5 minutes and is less than the minimal required value.

deploy/monitor/prometheus_rule.yaml
wei-lee: let's create some issues to capture future improvements.

camilamacedo86:
Hi @maleck13 @Nanyte25 please feel free to make suggestions for any part of the SOP's files. 
See here: https://github.com/aerogear/mobile-security-service-operator/tree/master/SOP. We will merge it, however, we can address your recommendations/suggestions in a new PR.

deploy/monitor/prometheus_rule.yaml
camilamacedo86: Done 👍 . However, after any change in the PR files, it is losing the approval. Please, could you do it again for we are able to merge it?

deploy/monitor/mss_prometheus_rule.yaml
wei-lee: @camilamacedo86 what about this comment?

camilamacedo86: It is true. Really tks to get it.

camilamacedo86: Done it + add further information in the SOPs as was requested/suggested by OPS.

camilamacedo86:
Hi @wei-lee I think we are able to move forward here now.

PR #155, opened 2019-07-25
Update Deployment GVK; fix CRD group; tighten permissions

grdryn:
## Motivation
JIRA: https://issues.jboss.org/browse/AEROGEAR-9616

- The operator ServiceAccount shouldn't be given permissions that it doesn't need.
- We shouldn't use a domain that we don't control for the group.
- We should use the latest version of the `Deployment` kind.

## What
This change was originally just intended to tighten the permissions, but I also took the opportunity to update to use the new version for `Deployment` (it took me a while to realise why `Deployment` wasn't working originally even when I had it specified in the role), and update the group part for the CRDs (we don't own `aerogear.com`, so that was likely just a typo originally).

## Why
Wouldn't that already be covered by the earlier "Motivation" section?

## How
See the "Files changed" tab, or navigate to https://github.com/aerogear/mobile-security-service-operator/pull/155/files in your browser.

## Verification Steps
- Check out this branch
- `sed -i -e "s|quay.io/aerogear/mobile-security-service-operator:master|quay.io/aerogear/mobile-security-service-operator:$(git rev-parse --short HEAD)|" deploy/operator.yaml`
- `make install`
- Make sure the service provisions correctly, and there are no error logs related to permissions in either the operator or service pods.
- Provision an app CR and make sure that works too, and there are no errors.

## Checklist:


deploy/cluster_role.yaml
```diff
@@ -3,64 +3,108 @@ kind: ClusterRole
...
   resources:
-  - deployments
-  - daemonsets
-  - replicasets
-  - statefulsets
+  - services
+  - persistentvolumeclaims
+  - events
+  - configmaps
+  - secrets
```
wei-lee: are `secrets` really required here?

deploy/cluster_role.yaml
```diff
@@ -3,64 +3,108 @@ kind: ClusterRole
...
+  resourceNames:
+  - mobile-security-service-operator
+  resources:
+  - deployments/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - pods
```
wei-lee: `pods` already listed above, probably can be removed from here?

deploy/cluster_role.yaml
grdryn: Yes, they're at least used in the `mobilesecurityservicebackup` controller.

deploy/cluster_role.yaml
grdryn: Ah yes, I'll remove this from here :+1:

aliok:
@grdryn The group is mentioned in MDC and MDC operator. I will create PRs for them unless you did it already.

aliok:
Verifying this now

grdryn:
Thanks for the reviews & verification folks!

camilamacedo86:
HI @grdryn @grdryn @wei-lee,

The changes performed in the rules here is causing issues in the monitoring feature and probably in the backup as well since the rules added regards this one was removed as well and these features were not tested in order to allow this merge.

In the following PR we are coming back the rules for we are able to do the release and if you wish @grdryn please track a JIRA for we are able to customize them properly after. Note that is the important highlight that both features need to be manually validated as well. Also, we should test all functionalities to be very sure that they are not affected by the changes too.  I hope that we can move forward in this way.

PR #157, opened 2019-07-30
fix(Prometheus Rules) : Remove MobileSecurityServicePodCount since it…

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9642

## What
- Remove MobileSecurityServicePodCount since it was considered redundant

## Why
According to the convention/standard, we have already the specific rule for each pod. 

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Remove MobileSecurityServicePodCount since it was considered duplicated [#157](https://github.com/aerogear/mobile-security-service-operator/pull/157)
 
 ## [0.3.0] - 2019-07-26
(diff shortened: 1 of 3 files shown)


laurafitzgerald:
@camilamacedo86 we should remove it from [here](https://github.com/integr8ly/installation/blob/master/roles/mobile_security_service/templates/mss_operator_prometheus_rule.yml.j2#L26) also. Creating a task for that: https://issues.jboss.org/browse/AEROGEAR-9645

camilamacedo86:
I think it should be another task @laurafitzgerald since will be required to perform a new release for each operator then we can remove from there as well. Could you track this need?  
IMHO: we should wait to see if no other changes will be required as well

laurafitzgerald:
@camilamacedo86 yes that's fine. There is a task linked above. Can you add that comment in the JIRA. The only reason I linked it here is to show that the related work is tracked.

camilamacedo86:
Hi @grdryn you approve the other tasks of the same JIRA. Could you also check this one for we are able to close the task?

PR #158, opened 2019-07-30
Fix nomenclature of CR/CRD's in MSS Opr

camilamacedo86:
## Motivation
- https://issues.jboss.org/browse/AEROGEAR-9653

## What
- Fixing nomenclature of CR/CRD's

## Why
The name of the files which are created/generated by the operator-sdk tools currently are no longer the same when the project started which was in a lower version. Note that we could just check it after adding the open-api command in the make code/gen. 

## How
- Rename files with the same name which the tool will generate them
- Doing the required adjustment in the make commands.

### IMPORTANT:
-  It should not affect the RHMI installation scripts since it is not using the make commands at all.
- Only the make commands are affected by this change

### Steps
1. Run `make install` and `make uninstall` and ensure that both commands worked as should be.(I mean check if it applied/delete the service+operator files in the terminal - just that)
2. Run `make example-app/apply` and `make example-app/delete` and ensure that both commands worked as should be. (I mean check if it  applied/delete the example app cr - just that)
3. Run `make backup/install` and `make backup/uninstall` and ensure that both commands worked as should be. (I mean check if it applied/delete the bkp cr - just that)
4. Just to ensure run `make install` and ensure that it will result in 3 pods running with success. 
5. Run `make code/gen` and ensure that no new files will be created at all.

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


camilamacedo86:
Closing it. We need check.fix it after.

PR #162, opened 2019-08-05
fix(operator metrics): path export metrics (AEROGEAR-9664)

camilamacedo86:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9664

## What
Fix MSS Monitor Service prometheus api path 

## Why
It was wrong

## Verification Steps
Apply this change in the Service Monitor in the testing cluster or check it changed here: https://prometheus-route-middleware-monitoring.apps.london-79f0.openshiftworkshop.com/alerts
Screenshot 2019-08-05 at 22 21 30

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/deploy/monitor/mss_service_monitor.yaml
+++ b/deploy/monitor/mss_service_monitor.yaml
@@ -7,7 +7,7 @@ metadata:
   name: mobile-security-service
 spec:
   endpoints:
-    - path: /api/metrics
+    - path: /rest/prometheus/metrics
       port: server
   selector:
     matchLabels:


laurafitzgerald:
👀

deploy/monitor/mss_service_monitor.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   name: mobile-security-service
 spec:
   endpoints:
-    - path: /api/metrics
+    - path: /rest/prometheus/metrics
```
laurafitzgerald: @camilamacedo86 I believe the existing route is the correct one. The Mobile Security Service exposes service metrics (such as MobileSecurityServiceApiHighRequestDuration) on the `/api/metrics` endpoint. We may need to add an additional one to collect the kube state metrics? but I believe we still need to have this one here.

camilamacedo86: HI @laurafitzgerald,

If you check after this change all started to work.  The kube state metrics is an issue for all projects and is related to Prometheus changes in integr8ly. See here : https://issues.jboss.org/browse/AEROGEAR-9704. I will ping you in the internal discussion in the chat as well.

camilamacedo86: Hi @laurafitzgerald after this change I could start to find the metrics expose there. 
I will do more tests to ensure.

camilamacedo86: I think you are correct @laurafitzgerald. So, I will close this one and commet in the jira. 
See here; https://github.com/aerogear/mobile-security-service/blob/4a7dd8b282592106fa9ed5c7836d1b48158dfaf1/api/swagger.yaml#L265

It is possible that it is something related to   https://issues.jboss.org/browse/INTLY-2810

PR #174, opened 2019-09-20
e2e test for MSS operator

STEPHINRACHEL:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9798

## What
Added e2e test for MSS operator

## How
Using Operator-SDK's test framework

## Verification Steps
Run  the following from repository

```shell
  operator-sdk test local ./test/e2e --go-test-flags "-v -parallel=2"
```

## Checklist:

## Progress

- [ ] Finished task
- [ ] TODO


psturc:
Also, renaming the `role*.yaml` files would break Makefile commands, for example [here](https://github.com/aerogear/mobile-security-service-operator/blob/master/Makefile#L39-L40). So you can either:
1. Fix Makefile
2. Rename `role*.yaml` files back and use [these parameters](https://github.com/aerogear/mobile-developer-console-operator/blob/master/Jenkinsfile#L123-L124) for running the tests in Jenkins instead. These parameters ensure that all manifest files (role, role_binding, service_account, etc.) and global manifest files (*_crd) get merged and used in test. So it's no longer a problem to have these files named differently.

psturc:
Build is failing because this operator requires a separate CR for deploying a DB for some reason..
https://github.com/aerogear/mobile-security-service-operator/blob/master/Makefile#L46

Also, I'm not sure if we should suddenly add all of the content from `vendor` folder, which is more than 7k files - @grdryn, wdyt? Should we leave it like that or "unify" it with our other operators, i.e. commit the vendor files?
If we don't decide to commit the `vendor` folder, there's an [additional step](https://github.com/aerogear/mobile-security-service-operator/blob/master/Makefile#L178) we need to add to the Jenkins pipeline which is required to run before running the tests, to ensure that all required dependencies are downloaded for the operator.

laurafitzgerald:
@psturc @STEPHINRACHEL we decided a while back to not include the vendor folder in this repo. You can see comments about it in thsi pr: https://github.com/aerogear/mobile-security-service-operator/pull/87

grdryn:
I don't have a problem with committing the vendor directory; but if it's being done, consider adding the `unused-packages` prune option like I've done here: https://github.com/aerogear/unifiedpush-operator/pull/72/files#diff-836546cc53507f6b2d581088903b1785R59

That should drastically reduce the number of files committed. The only thing to think of is that when importing a new package (i.e. one that we haven't used before) from an existing dependency, we'd need to run `dep ensure` again to get it pulled in.

test/e2e/mobile_security_service_test.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		t.Fatalf("Failed to initialize DeploymentConfig Client : %v", err)
+	}
+
+	//Delete MSS database CR
+	if err := deleteMSSdbCustomResource(t, f, ctx, mssDBTestCR, namespace, mssdbName); err != nil {
+		t.Fatal(err)
+	}
+	//Ensure MSS database was deleted successfully
+	if err = waitForDeploymentConfig(t, *dcV1Client, namespace, mssdbName, 0); err != nil {
```
psturc: Should we use `e2eutil.WaitForOperatorDeployment` instead since MSS doesn't use deploymentConfigs?

test/e2e/mobile_security_service_test.go
```diff
@@ -0,0 +1,244 @@
...
+		t.Fatal(err)
+	}
+	t.Log("MSS database was deleted successfully")
+
+	//Delete MSS CR
+	if err := deleteMSSCustomResource(t, f, ctx, mssTestCR, namespace, mssName); err != nil {
+		t.Fatal(err)
+	}
+	//Ensure MSS was deleted successfully
+	if err = waitForDeploymentConfig(t, *dcV1Client, namespace, mssName, 0); err != nil {
```
psturc: Same here, see previous comment

test/e2e/mobile_security_service_test.go
STEPHINRACHEL: I used it since if(replicas == 0) condition is not checked in waitForDeployment() method  in e2eutil for testing deletion.

psturc: Ok in that case, let's skip this step then, because checking for DC doesn't make sense here.

PR #177, opened 2019-10-04
AEROGEAR-9985 - Upgrade mss-operator to operator-sdk v0.10.0

aliok:
Verification instructions:

Enable anyuid:
```
minishift addons enable anyuid
minishift addons apply anyuid
```

Install Integreatly application monitoring operator:
```
clone https://github.com/integr8ly/application-monitoring-operator/ under your GOPATH.
make cluster/install
```

If you see an error in Prometheus regarding file system permissions, you need to `chmod 777` the host path:
```
oc get pvc
oc get pv pv0058 -o yaml
# host path is /var/lib/minishift/base/openshift.local.pv/pv0058
minishift ssh
sudo chmod -R 777 /var/lib/minishift/base/openshift.local.pv/pv0058
exit
# delete prometheus pod to have the statefulset recreate it
```

Deploy MSS operator, monitoring for it, and an example app
```
# first, change image in operator.yaml to aliok/mss_opr_oct_1_1337
make install
make monitoring/install
oc project mobile-security-service-apps
make example-app/apply
```

Check these:
1. MSS is running, you can see the example app in MSS console
2. You can see MSS operator and MSS service in Prometheus targets
3. You can see MSS operator and MSS service related rules in Prometheus rules
4. You can see MSS operator and MSS service related dashboards in Grafana

Note that some of the stuff in Prom/Grafana doesn't show up properly in Minishift. I assume they will show up in RHMI cluster as these Prom queries came all the way until today.


aliok:
@grdryn @psturc @laurafitzgerald 
Ready for review now

.circleci/config.yml
```diff
@@ -7,7 +7,9 @@ jobs:
     working_directory: /go/src/github.com/aerogear/mobile-security-service-operator
 
     docker:
-      - image: circleci/golang:1.10
+      - image: circleci/golang:1.12
```
grdryn: @psturc Do you know what version of Go is installed in the Jenkins agent that compiles?

Makefile
```diff
@@ -216,13 +214,13 @@ code/gen:
 .PHONY: test/run
 test/run:
 	@echo Running tests:
-	GOCACHE=off go test -cover $(TEST_PKGS)
```
grdryn: What was the significance of this being here originally, and is it good that it's being removed now?

cmd/manager/main.go
```diff
@@ -91,10 +102,41 @@ func main() {
...
+	}
+
+	err = createServiceMonitor(cfg, operatorNamespace, service)
+	if err != nil {
+		log.Info("Could not create ServiceMonitor object", "error", err.Error())
+		// If this operator is deployed to a cluster without the prometheus-operator running, it will return
+		// ErrServiceMonitorNotPresent, which can be used to safely skip ServiceMonitor creation.
+		if err == metrics.ErrServiceMonitorNotPresent {
+			log.Info("Install prometheus-operator in you cluster to create ServiceMonitor objects", "error", err.Error())
+		}
```
grdryn: Could you wrap this in a conditional block so that it only gets executed if the service gets created successfully? This is something I originally forgot to do in the other ones (there are open PRs now), and without it the operator isn't runnable locally. See this commit in my PR for the MDC operator:

https://github.com/aerogear/mobile-developer-console-operator/pull/60/commits/0663b011bb58fc91e49f20f8c3fd35d2980f2cbd

Essentially:
```go
if service != nil {
    ...
}
```

deploy/role.yaml
```diff
@@ -107,3 +107,24 @@ rules:
   - namespaces
   verbs:
   - get
+
+# monitoring related perrmissions
```
grdryn:
```suggestion
# monitoring related permissions
```

grdryn:
I haven't done any verification yet, I will though :+1:

grdryn:
@aliok along with the other comments I left in my original review, could you also add a stage like the following to the Jenkinsfile, so that we're using the correct version of the operator-sdk CLI there: https://github.com/aerogear/unifiedpush-operator/blob/master/Jenkinsfile#L61..L73

.circleci/config.yml
aliok: @grdryn 
Things were failing badly with the SDK with that version of Golang. Some dependency required Golang 1.12.

@psturc Can we update Golang version in Jenkins, if it is old?

Makefile
aliok: From Go 1.1.2 release notes (https://golang.org/doc/go1.12)

> The build cache is now required as a step toward eliminating $GOPATH/pkg. Setting the environment variable GOCACHE=off will cause go commands that write to the cache to fail.

I had to remove this parameter to make things working with Go 1.12

cmd/manager/main.go
aliok: Cool. Thanks for the heads up!

.circleci/config.yml
psturc: the version of go in Jenkins slave is [v1.12.6](https://github.com/RHCloudServices/centralci/blob/master/ansible/roles/go/defaults/main.yml#L3)

aliok: Cool, should be no problems then!

aliok:
All comments addressed now.
Merging!

PR #179, opened 2019-11-13
ci: publish dev image pointing to master tag of operand

jhellar:
## Description

https://issues.jboss.org/browse/AGMS-1135

Added stage to Jenkinsfile to rebuild container image pointing to master of its operand.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -180,6 +180,46 @@ pipeline {
                 )
             }
         }
+        // rebuild and push the image so that it points to master tag of operand (for nightly testing purposes)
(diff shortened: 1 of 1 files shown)


grdryn:
@jhellar Is this needed in this case? Note that here you can set the operand image through the CR: https://github.com/aerogear/mobile-security-service-operator/blob/master/deploy/crds/mobile-security-service_v1alpha1_mobilesecurityservice_cr.yaml#L40

jhellar:
@grdryn did not notice this. Thanks, will use the CR.

